PR #13, opened 2016-05-10
phpfpm container renamed to php

kusmierz:
The phpfpm container has also phpcli, so maybe it would be better to rename it to just `php`.

Then you can simply run

```
docker-compose exec php bin/magento ...
```

(as of docker-compose 1.7.x).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@
 app:
   image: mageinferno/magento2-nginx:1.9.14-0
   links:
-    - phpfpm
+    - php:phpfpm
     - db
   volumes_from:
     - appdata
@@ -18,7 +18,7 @@ appdata:
 #    - ~/.composer:/var/www/.composer
 #    - ./www/app/code:/srv/www/app/code
 
-phpfpm:
+php:
   image: mageinferno/magento2-php:7.0.5-fpm-0
   links:
     - db
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This file is provided as an example development environment using Mage Inferno M
 app:
   image: mageinferno/magento2-nginx:1.9.14-0
   links:
-    - phpfpm
+    - php:phpfpm
     - db
   volumes_from:
     - appdata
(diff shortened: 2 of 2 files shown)


markoshust:
I don't want to rename this, because the base image is built from `phpfpm`. It's possible that in the future we will create image(s) from the `php` image, and then we would be in trouble.

markoshust:
You know what, I'm going to disagree with myself :) Since this is just in the docker-compose, I think it does make sense to rename this to php to keep things simple. We are already doing this with the database layer in naming it `db`.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@
 app:
   image: mageinferno/magento2-nginx:1.9.14-0
   links:
-    - phpfpm
+    - php:phpfpm
```
markoshust: can you name this just `php`.

kusmierz:
You can also call `db` container as `mysql`/`mariadb` and alias it as `db` for the `app` :)

and about `phpfpm` vs `php`... you named `app` the "...-nginx..." container, so I don't think this is an issue. But it would be easier to call, as I wrote above :)

docker-compose.yml
kusmierz: I could, but why would you like to change it? It requires more changes in configuration file(s) (see `mageinferno/magento2-nginx` image). For example I could add `ENV PHP_HOST` or make ugly change default host in [mageinferno/docker-magento2-nginx/Dockerfile#L4](https://github.com/mageinferno/docker-magento2-nginx/blob/master/Dockerfile#L4).

markoshust:
Let's think about this for a while. I think we should probably avoid renaming everything at this point, because it might cause some confusion. I understand the aliases here, but I think its simpler to not use aliases at all. Since nginx config at https://github.com/mageinferno/docker-magento2-nginx/blob/master/Dockerfile references `phpfpm`, we should probably just keep that naming convention for now. 

I understand this approach isn't used for nginx (`app`) and mariadb (`db`), however neither of those has a dependency such as `phpfpm` within their configurations, so we should probably keep the current approach going forward.

PR #35, opened 2016-11-13
Added Magento Enterprise environment variable to toggle support for Enterprise installations using Composer

careysizer:
See [related PR 31](https://github.com/mageinferno/docker-magento2-php/pull/31) on docker-magento2-php which adds support for Magento Enterprise in the php image.


markoshust:
Another great update, thanks mucho

PR #51, opened 2017-02-09
Fixed setup image version, Enterprise installation support

careys7:
Enterprise installation wasn't working using the 7.0 image. 

The setup script included in that image doesn't support the `M2SETUP_USE_COMPOSER_ENTERPRISE` variable.

Looking at the image used for phpfpm, it seems like the setup one should have been bumped too?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ dbdata:
     - /var/lib/mysql
 
 setup:
-  image: mageinferno/magento2-php:7.0-fpm-0
+  image: mageinferno/magento2-php:7.1-fpm-0
   command: /usr/local/bin/mage-setup
   links:
     - db


markoshust:
@careys7 thank you sir for your updates

careys7:
@markoshust - apologies - I was just about to cancel this PR as it needs more work:

http://devdocs.magento.com/guides/v2.0/install-gde/system-requirements-tech.html#php

> Not supported:
> PHP 7.1
> PHP 5.4

markoshust:
@careys7 i find it hard to believe that 7.0 is supported... but 7.1 is not? 🙃

careys7:
Yes I know 👎 It looks like there is a few small incompatibles: https://github.com/magento/magento2/issues/7663

Composer now fails on the PHP requirement during installation for Enterprise, and I assume Community is the same. I should have waited for this to complete before opening PR. 

Possibly the best option is to add the `M2SETUP_USE_COMPOSER_ENTERPRISE` flag into a PHP70 image for the time being?

PR #53, opened 2017-02-09
Revert "Fixed setup image version, Enterprise installation support"

careys7:
See https://github.com/mageinferno/magento2-docker-compose/pull/51

This reverts commit c1ca67d6515928156277ed570369fca4c329033b.


denisinla:
I can confirm that 7.1 install(spent an hour trying to figure out what was wrong with the install) is broken on CE edition as well. Reverting back to 7.0.

PR #132, opened 2019-02-22
allow setup without ssh credentials

ifiokjr:
Currently, the installation process fails when a user hasn't yet set up their GitHub account or an SSH public key.

Using the `https://` URL prevents things from failing.


markshust:
thanks, i can't really see any downsides of this approach since 99% of people will use this for read-only anyways. merged in 👍

PR #149, opened 2019-05-12
add mailhog

torhoehn:
I added mailhog for development environments, so mails will not send directly to real mail servers.
If you accept this PR, I could edit the README.

### DIFF
--- a/compose/magento-2/docker-compose.dev.yml
+++ b/compose/magento-2/docker-compose.dev.yml
@@ -26,3 +26,9 @@ services:
   # Disabling cron by default as it uses higher CPU, enable if needed
   #cron:
   #  volumes: *appvolumes
+
+  mailhog:
+    image: mailhog/mailhog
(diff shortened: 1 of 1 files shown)


markshust:
@torhoehn do you have documentation of how to use mailhog? i've never used it. i took a quick glance and since mailhog is a self-hosted project this might work really well.

torhoehn:
@markshust If you start it as it is, just open your domain e.g. magento2.test:8080. Als mails send in Magento will be catched and you'll have a button to release it "in the real world". A popup will appear to insert credentials for an SMTP server.

markshust:
@torhoehn I added the mhsendmail php package along with mailhog config for php.ini (this won't work without this).i tested locally and it appears to work. all merged in and tagged, should be ready when next build is complete in the `php:7.2-fpm-2` and `php:7.1-fpm-11` tags

PR #150, opened 2019-05-15
Use container id for docker commands

jerrylopez:
> FYI, I came across this issue using the [mage2click/docker-magento-mutagen](https://github.com/mage2click/docker-magento-mutagen) project which is based off of this project. I submitted the same PR there: https://github.com/mage2click/docker-magento-mutagen/pull/20

When using the `bin` scripts I noticed that sometimes the commands worked and sometimes they would throw an error like this: 
```
Error: No such container:path: magento-231_phpfpm_1:/var/www/html/auth.json
```

In this situation, my directory name is `docker-magento-231` and the container's name is `docker-magento-231_phpfpm_1`. Weirdly enough anything from the first `-` and before is being omitted from the docker commands. I found that the width of the terminal window was actually affecting this. 

**Example**
If the window was wide enough the docker commands run just fine because the container name is correct

image

If the window is too narrow the container name is trimmed because the docker-compose name output is multi-lined

image

I found that the issue lies when grep is used to retrieve the container name. I tested out `grep` in a different scenario and it doesn't seem to be the cause of the issue. A co-worker actually mentioned that this could have something to do with `docker-compose` and how the data is piped.

I didn't dig too much into the situation because I went the route of using the container id instead for the `docker` commands.

### DIFF
--- a/compose/magento-2-windows/bin/copydir.ps1
+++ b/compose/magento-2-windows/bin/copydir.ps1
@@ -3,4 +3,4 @@ Param(
   [Parameter(Mandatory=$True,Position=1)]
   [string]$dir
 )
-docker cp $(docker-compose ps|grep phpfpm|awk '{print $dir}'):/var/www/html/$dir src/
+docker cp $(docker-compose ps -q phpfpm|awk '{print $dir}'):/var/www/html/$dir src/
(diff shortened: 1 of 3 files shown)


markshust:
@jerrylopez thank you very much! i noticed this myself in testing but wasn't sure what was causing it (never thought of a line break). nice fix, all merged

jerrylopez:
My pleasure! 😄 it was a super weird issue

PR #171, opened 2019-08-09
Adds sockets PHP extension

mark-gene:


markshust:
Thanks for the contribution, however I noticed a lot of other PHP modules were missing and excluded from the install. This was added a while ago in https://github.com/markshust/docker-magento/commit/4432033d6142172f0ceb374abc589d82679596dc - so going to close out this PR.

PR #173, opened 2019-08-12
Adding ioncube loader to the container

emberdyn:
This is a great tool.  Wondering if we can add ioncube loader to the container?  

Thanks.


emberdyn:
@markshust  - Is it a possibility to add this in so that we can use ioncube for local development?

images/php/7.2/Dockerfile
```diff
@@ -44,6 +44,20 @@ RUN docker-php-ext-install \
   xsl \
   zip
 
+RUN cd /tmp \
```
markshust: Can you keep just this single `RUN` command and remove all of the other updates? We don't want to add backports to the repo to add libsodium, it's already added in libsodium-dev above.

images/php/7.1/Dockerfile
```diff
@@ -44,6 +44,22 @@ RUN docker-php-ext-install \
   xsl \
   zip
 
+
```
markshust: Let's remove this extra blank line, keep just the `RUN` here and remove all of the other updates

images/php/7.2/Dockerfile
emberdyn: Whoops.  My bad.  Thanks for catching that.  I have made the requested changes.  Thanks!

images/php/7.1/Dockerfile
emberdyn: I have made the requested changes.  Thank you!

markshust:
Great, all merged! I'll rebuild a new set of images shortly.

avra911:
@markshust , @emberdyn , please see my PR: https://github.com/markshust/docker-magento/pull/219.

Thank you!

markshust:
Great, all merged. I'll see what I can do about getting another build out shortly.

PR #192, opened 2019-11-20
Fix Magento2 setup script with n98-magerun2.phar

rataj86:
I fixed setup script where is used n98-magerun2 without .phar extension.

In bin/n98-magerun2 file is call:
#!/bin/bash
bin/cli bin/n98-magerun2.phar "$@"

and it didn't work when we had saved magerun script: src/bin/n98-magerun


markshust:
Thanks @rataj86, merged!

PR #196, opened 2019-11-26
Allow 'src' being a symlink to actual directory of Magento project on disk

Seb2nim:
The fix concerns both `bin/copyfromcontainer` and `bin/copytocontainer` to allow 'src' being a symlink.
Actually, `docker cp` fails on copying files on symlink.
The fix consists in resolving "real path" to 'src' to make `docker cp`.
I choose `cd + pwd` for Mac OSX compliance.


Seb2nim:
Sorry, gitignore inside, starting again

PR #197, opened 2019-11-26
Fix dev-urn-catalog-generate

Seb2nim:
Fix dev-urn-catalog-generate :
Restore the sed replace and made it globally (some <resource> are multiline)
Works if src is a symlink


markshust:
Thanks for this! I didn't test, but trust this works as intended and will merge in.

PR #212, opened 2020-01-21
Added rabbitmq support

rmsundar1:
Added RabbitMq support

> To access RabbitMq Management UI - http://localhost:15672


markshust:
Thanks @rmsundar1! This passes my review. I'll just test locally, if all looks good I'll merge in.

PR #221, opened 2020-02-08
Make sure directory name is converted to lowercase

shochdoerfer:
Since docker uses lower case volume names, make sure to convert the current directory to lowercase before passing it to the docker command.


markshust:
Thanks, all merged!

PR #225, opened 2020-02-17
Add ioncube for php 7.2 and 7.3 Dockerfile

rangerz:
For php7.2 this path of php-ext is changed: 
/usr/local/lib/php/extensions/no-debug-non-zts-20151012/ioncube_loader_lin_7.2.so
/usr/local/lib/php/extensions/no-debug-non-zts-2017XXXX/ioncube_loader_lin_7.2.so
Also add ioncube for php 7.3

Its commit will work normally when path changes again in the future.

### DIFF
--- a/images/php/7.2/Dockerfile
+++ b/images/php/7.2/Dockerfile
@@ -44,12 +44,12 @@ RUN docker-php-ext-install \
   xsl \
   zip
 
-RUN cd /tmp \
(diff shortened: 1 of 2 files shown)


rangerz:
I got a better solution in #226

markshust:
great, thanks yes #226 looks really nice. all merged in. thanks!

PR #232, opened 2020-03-03
Updated mysql, n98-magerun2, and setup

rangerz:
1. Updated README.md for new `bin/mysql`
2. Improved `bin/mysql`, `bin/setup` and `bin/98-magerun2`

### DIFF
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -9,10 +9,6 @@ sed -e 's/5f5929ef9f2ec4ca048a2add261d22c92807630f/ce31e720d60451784b9fdb3769e43
 docker-compose -f docker-compose.yml up -d
 sleep 1 #Ensure containers are started...
 
-curl -O https://files.magerun.net/n98-magerun2.phar
-chmod +x n98-magerun2.phar
-mv n98-magerun2.phar src/bin/n98-magerun2.phar
-
 echo "Copying all files from host to container..."
 rm -rf src/vendor #Clear for step below
 bin/copytocontainer --all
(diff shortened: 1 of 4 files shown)


compose/bin/setup
```diff
@@ -44,16 +40,16 @@ if [ -z "$PUBLIC_KEY" ] || [ -z "$PRIVATE_KEY" ]; then
 fi
 
 if [ -n "$PUBLIC_KEY" ] && [ -n "$PRIVATE_KEY" ]; then
-    bin/clinotty composer config http-basic.repo.magento.com $PUBLIC_KEY $PRIVATE_KEY
+    bin/composer config http-basic.repo.magento.com $PUBLIC_KEY $PRIVATE_KEY
```
markshust: Automated installs require the use of the `-T` flag when calling exec with docker compose, otherwise they will fail. I'd suggest to change the `bin/composer` script to use `bin/clinotty` instead of `bin/cli`, however it adds some nice pretty colors when running `bin/composer` that we'll lose if we switch. So, I think we should back out any changes in this `bin/setup` script that involved removing the `bin/clinotty` command.

rangerz:
@markshust I try to update them. 
But I still remove n98-magerun2 install steps in `bin/setup`.  I think `bin/n98-magerun2` can cover well for the steps.

compose/bin/setup
rangerz: @markshust  Thanks for your suggestions

rangerz:
Let me change this pull request form branch

PR #235, opened 2020-03-04
Updated mysql, n98-magerun2, and setup

rangerz:
@markshust  It's the same with #232 
The mechanism of Pull Request is not friendly to me.


markshust:
I made a few updates. The updates to the php.ini file could have caused issues with Magento's performance, specifically the realpath directives, so I just backed them out.

In the future, you can just continue pushing to the branch, the PR will be automatically updated! Took me a while to find that out. I'm not fond of GitHub's code review process either.

Cheers -Mark

markshust:
I’m pretty sure that blog post is incorrect - an “Invalid form key” error
shows when the code is not properly passing around the hidden form key in
Magento.

See
https://github.com/magento/magento2/issues/19712#issuecomment-446557313

rangerz:
@markshust 
This issue happened to my website on edit product page, and the change can solve this issue.
I think this issue won't happen to the general product attribute set. ex: official sample data, luma.
In my website case, there are 1986 input vars for a saving POST request in edit product. 
After my research, I found "Customizable Options" will increase the amount of variables. 
ex: **ONE Option Value** for Customizable Options(Type: Radio Buttons)
```
product[options][16][values][0][option_type_id]: 123
product[options][16][values][0][option_id]: 321
product[options][16][values][0][sku]: 
product[options][16][values][0][sort_order]: 0
product[options][16][values][0][default_title]: xxx
product[options][16][values][0][store_title]: xxx
product[options][16][values][0][title]: xxx
product[options][16][values][0][default_price]: 0.000000
product[options][16][values][0][default_price_type]: fixed
product[options][16][values][0][store_price]: 0.000000
product[options][16][values][0][store_price_type]: fixed
product[options][16][values][0][price]: 0.00
product[options][16][values][0][price_type]: fixed
product[options][16][values][0][is_use_default]: false
product[options][16][values][0][initialize]: true
```
And the form_key will be in the end of POST request.
So we just get the error message for "Invalid Form Key. Please refresh the page"

markshust:
@rangerz did you test this change out locally first to see if it resolves your issue?

rangerz:
@markshust Yes, I tested on my local site. After the change `max_input_vars = 10000`, it can resolves my issue, `Invalid Form Key. Please refresh the page` when save product in M2 backend.

PR #236, opened 2020-03-10
Updated bin/remove and bin/status

rangerz:
Updated `bin/remove` to include `bin/removevolumes` function, and with `--all` can remove all images, containers, networks, and volumes
Also update README.md

Fixed issues: 
1. `docker-compose rm` only remove stoped containers. It should add `--stop`, if required, before removing.
2. `bin/removevolumes`, for volume naming rule need to use lowercase and skip dot `.` characters, and volume list should include rabbitmqdata, ssldata, and appvolumes
3. `bin/status` should appear mailhog status (define in `docker-compose.dev.yml`)

Because English is not my native language, please help me update these description if necessary.
Thanks.

PR #237, opened 2020-03-11
Added bin/removeall and fixed few issues

rangerz:
After discussion #236 and modified this pull
1. Added `bin/removeall` for removing all containers, networks, volumes, and images.
2. Fixed docker-compose issue, it should include docker-compose.yml and docker-compose.dev.yml 
3. `bin/remove`: Add `--stop` param that stop containers first and remove stoped containers
4. Keep `bin/removevolumes` and fixed volume naming rule.
5. Updated README.md

PS: 
Because `bin/removevolumes` action is removing volume only, it will need to maintain in the future when volume setting changing in docker-compose.yml.
The `docker-compose down` is inappropriate because it [default action](https://docs.docker.com/compose/reference/down/) is removing containers and networks


markshust:
@rangerz very nice work, all merged!

PR #251, opened 2020-04-17
Fix dev-urn-catalog-generate

gander:
Sed manual: `-i[SUFFIX], --in-place[=SUFFIX]`
Remove `misc.xml` from container

### DIFF
--- a/compose/bin/dev-urn-catalog-generate
+++ b/compose/bin/dev-urn-catalog-generate
@@ -5,7 +5,7 @@ bin/clinotty rm misc.xml
 
 REAL_SRC=$(cd -P "src" && pwd)
 
-sed --in-place=.bak "s?/var/www/html?$PROJECT_DIR$/src?g" $REAL_SRC/misc.xml
+sed -i .bak "s?/var/www/html?$PROJECT_DIR$/src?g" $REAL_SRC/misc.xml
 rm $REAL_SRC/misc.xml.bak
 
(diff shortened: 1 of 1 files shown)


markshust:
@gander mac doesn't seem to have an `--in-place` argument, and it's a lot more common to use `-i`, so kept that flag as-is.

I'll merge in removing the unneeded misc.xml from the container though.

thanks!

gander:
You make a mistake all the time. There is no space after `-i`! Check the documentation. I proposed a long name to make it more obvious.

https://linux.die.net/man/1/sed

gander:
If the syntax of this command is different in mac and linux after all, you may need to distinguish in what system it is running?

gander:
@markshust https://stackoverflow.com/a/4247319/4559414

markshust:
I didn't make a mistake -- `-iCOMMAND` doesnt work on OSX, and neither does `--in-place` (that flag doesn't exist on OSX). it appears the `-i COMMAND` format only doesn't work on the GNU o/s, which I'm totally ok not supporting. I've always used `-i COMMAND` my entire life, so i know it works on almost all platforms.

gander:
I understand that. What are you going to do about it? I suggest either detecting the system and running this command differently depending on it, or stop using this switch.

markshust:
I'm going to live with it not working on GNU :) This is an acceptable bug for me to pass on. If someone else fixes it, I would gladly take in a PR for it.

PR #269, opened 2020-05-15
189: add the basic mftf setup information

### LINKED ISSUE #189: How to run mftf using docker-magento?
Hello dearest docker-magento community,

I've been trying to give running MFTF tests on the instance running inside docker a  shot, to no success thus far. I'm following the [MFTF Getting Started guide](https://devdocs.magento.com/mftf/docs/getting-started.html#run-tests) but I get stumped on a seemingly simple issue: having the test generation bail because it can't reach the magento storefront (over http or https) from inside the container. I get (after exposing some of the URL information a bit  in the error message myself):

```
➜ ./bin/clinotty vendor/bin/mftf generate:tests


In ModuleResolver.php line 444:

  Could not retrieve API token from Magento Instance (Could not find Magento
  Backend Instance at MAGENTO_BACKEND_BASE_URL or MAGENTO_BASE_URL (http://ma
  gento2.test/rest/V1/integration/admin/token))
```

Turns out, I don't know if I can reach the magento storefront from inside the container, regardless of what domain or IP I try to use. 

This is probably because I don't understand how docker-compose works, but figured I'd ask here as someone possibly does know! I think it'd be cool if we could run MFTF tests from inside the container - I think it should be theoretically doable.

dmanners:
- Update the readme with mftf setup instructions,
 - create a wrapper bin/mftf command
 - update the docker-compose and docker-compose.dev yml files with the needed docker images

Fixes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,7 @@ You'll now have an updated `bin/update` helper script, and can run it to update 
 - `bin/fixperms`: This will fix filesystem permissions within the container.
 - `bin/grunt`: Run the grunt binary. Ex. `bin/grunt exec`
 - `bin/magento`: Run the Magento CLI. Ex: `bin/magento cache:flush`
+- `bin/mftf`: Run the Magento MFTF. Ex: `bin/mftf build:project`
 - `bin/mysql`: Run the MySQL CLI with database config from env/db.env. Ex `bin/mysql -e "EXPLAIN core_config_data"`
 - `bin/n98-magerun2`: Access the n98 magerun CLI. Ex: `bin/n98-magerun2 dev:console`
 - `bin/node`: Run the node binary. Ex. `bin/node --version`
@@ -413,6 +414,33 @@ Next, open up the `bin/start` helper script and uncomment the line:
 ```
 
 Finally, restart the containers with `bin/restart`. After doing so, everything is now configured and you can use a browser extension to profile your Magento store with Blackfire.
+
+### MFTF
+
+To work with MFTF you will need to first enable the `selenium` image in the `docker-compose.dev.yml` file.selenium
+Then you will need to run the following.
+
+1. Run mftf build process `bin/mftf build:project`. This should build the basic setup for mftf in your project,
+1. Update the extra host values to match your Magneto URL and IP in `docker-compose.dev.yml`,
+1. Update the values in `src/dev/tests/acceptance/.env`. Including adding the new line `SELENIUM_HOST=selenium` to define for Codeception which host to connect to,
+1. Run a sample test `bin/mftf run:test AdminLoginTest`
+1. Update your nginx.conf to allow access to the dev section with the following before the final deny all section.
+
+```
+location ~* ^/dev/tests/acceptance/utils($|/) {
+    root $MAGE_ROOT;
+    location ~ ^/dev/tests/acceptance/utils/command.php {
+        fastcgi_pass   fastcgi_backend;
+        fastcgi_index  index.php;
+        fastcgi_param  SCRIPT_FILENAME  $document_root$fastcgi_script_name;
+        include        fastcgi_params;
+    }
+}
+```
+
+For debugging you can connect to the selenium image using a VCN client.
+Connect with the VCN option and `127.0.0.1:5900`
+You can also run `bin/mftf doctor` to validate all sections are setup correctly.
 
 ## Credits
 
--- a/compose/bin/mftf
+++ b/compose/bin/mftf
@@ -1,0 +1,2 @@
+#!/bin/bash
+bin/clinotty vendor/bin/mftf "$@"
--- a/compose/docker-compose.dev.yml
+++ b/compose/docker-compose.dev.yml
@@ -39,3 +39,15 @@ services:
   # Disabling cron by default as it uses higher CPU, enable if needed
   #cron:
   #  volumes: *appvolumes
+
+  # Disabling selenium by default as it is not required for all use cases
+  #selenium:
+  #  image: selenium/standalone-chrome-debug:3.8.1
+  #  ports:
+  #    - "5900:5900"
+  # Uncomment two lines below & replace IP with result of: docker run --rm alpine ip route | awk 'NR==1 {print $3}'
+  # Replace the "magento2-composer.loc" with the url of your site
+  #  extra_hosts:
+  #    - "magento2.test:172.17.0.1"
+  #  links:
+  #    - app
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -63,6 +63,9 @@ services:
   #    - db
   #  volumes: *appvolumes
 
+  #selenium:
+  #  image: selenium/standalone-chrome-debug:3.8.1
+
 volumes:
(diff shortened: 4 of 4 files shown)


dmanners:
Hi @markshust This is the setup we have locally. Not sure if we need all this or maybe a little more for a good generic setup.

Questions that I have are:

1. Do we need the value in the docker-compose.yml or just the dev one? I'm not too great with docker so have not had it set up,
1. Should we change it to have a docker-compose.selenium.yml similar to the warden set up so you only start these when needed for example on a bin/start-mftf command or something,
1. Do we need the link between selenium and app images? Adding the extra hosts seems to configure the `/etc/hosts` file on the selenium image which should be enough to make sure it can connect back via the browser,
1. Would it make sense to add the nginx conf needed by default rather than just document the addition needed for the command style tests?

Happy to help out if you have any more ideas/change requests.

markshust:
Thanks @dmanners! This looks pretty straight-forward. I'm going to test and review this over the next couple days so we can hopefully merge this in really soon.

dmanners:
Sounds good, it could be that I missed something from our project's docker-composer.yml file so if something is missing or not working then feel free to drop me a message and I can try dig into the files.

markshust:
Hi David, I’ll review this and try to get it merged in today. Cheers -Mark

markshust:
Hi @dmanners,

In testing, I received the following output when running the sample test:

```
magento235$ bin/mftf run:test AdminLoginTest


 ! [NOTE] MFTF NOTICES:                                                         
 !        DEPRECATED ACTIONS: "executeInSelenium" and "performOn" actions will  
 !        be removed in MFTF 3.0.0                                              
 !        DEPRECATED TEST PATH: support for                                     
 !        "dev/tests/acceptance/tests/functional/Magento/FunctionalTest will be 
 !        removed in MFTF 3.0.0                                                 
 !        XSD schema change to only allow single entity per xml file for all    
 !        entities except data and metadata in MFTF 3.0.0                       
 !                                                                              


In ModuleResolver.php line 758:
                                                                               
  Could not retrieve API token from Magento Instance (Could not find Magento   
  Backend Instance at MAGENTO_BACKEND_BASE_URL or MAGENTO_BASE_URL)            
                                                                               

run:test [-k|--skip-generate] [-r|--remove] [-f|--force] [-a|--allow-skipped] [-d|--debug [DEBUG]] [--] <name>...
```

I'm guessing this inability to connect to the Magento instance has something to do with Docker not being routed correctly to the right container. I can go deeper in this, but wanted to check first if you knew an easy solution. The `.env` file I'm using in my dev test env is:

```
MAGENTO_BASE_URL=https://magento235.test/
MAGENTO_BACKEND_NAME=admin_2ik883
MAGENTO_ADMIN_USERNAME=john.smith
MAGENTO_ADMIN_PASSWORD=password123
BROWSER=chrome
MODULE_WHITELIST=Magento_Framework,ConfigurableProductWishlist,ConfigurableProductCatalogSearch
ELASTICSEARCH_VERSION=7
SELENIUM_HOST=selenium
```

dmanners:
Hmm yeah sounds like it cannot connect to your other containers.

For me I was able to do this by setting up the `extra_hosts` option against the selenium image so that it could connect back. For you I guess this should be:

```
- "magento235.test:172.17.0.1"
```

This should update the `/etc/hosts` file in the selenium container. I was able to validate this by using a client to connect to the container and then check the following:

1. That the right ip/url is in the `/etc/hosts/ files,
1. That a browser can actually connect to your host,

markshust:
> Hi @markshust This is the setup we have locally. Not sure if we need all this or maybe a little more for a good generic setup.
> 
> Questions that I have are:
> 
> 1. Do we need the value in the docker-compose.yml or just the dev one? I'm not too great with docker so have not had it set up,

You would just need the entry in the `.dev` file, running `bin/start` will merge these two files together upon start. The containers only need to be in both places if you are sharing filesystem mounts.

> 2. Should we change it to have a docker-compose.selenium.yml similar to the warden set up so you only start these when needed for example on a bin/start-mftf command or something,

This is an interesting idea. I've had the `cron` service also in the main `docker-compose.yml` file but commented out, and always found this ugly. The format is fine for now but I'll think through this a bit -- I think having separate files for this makes sense, if at least to get rid of all of those commented out lines.

> 3. Do we need the link between selenium and app images? Adding the extra hosts seems to configure the `/etc/hosts` file on the selenium image which should be enough to make sure it can connect back via the browser,

Selenium is mounted back to the host on port 5900, so we should be able to connect directly to the selenium container by using `mydomain.test:5900` without any additional `/etc/hosts` entries.

> 4. Would it make sense to add the nginx conf needed by default rather than just document the addition needed for the command style tests?

I think this does make a lot of sense to add it by default for an easier setup. I can add this in once I can confirm everything is working in the current setup.

> 
> Happy to help out if you have any more ideas/change requests.

Thanks so much! This addition is awesome, I've had a lot of requests for this fairly recently. This will make this setup so much better.

markshust:
> Hmm yeah sounds like it cannot connect to your other containers.
> 
> For me I was able to do this by setting up the `extra_hosts` option against the selenium image so that it could connect back. For you I guess this should be:
> 
> ```
> - "magento235.test:172.17.0.1"
> ```
> 
> This should update the `/etc/hosts` file in the selenium container. I was able to validate this by using a client to connect to the container and then check the following:
> 
> 1. That the right ip/url is in the `/etc/hosts/ files,
> 2. That a browser can actually connect to your host,

Ah I didn't uncomment that before running `bin/mftf`. I'm retesting.

markshust:
Ok I managed to get to this:

```
magento235$ bin/mftf doctor


 Requesting API token for admin user through cURL ...

 [ERROR] Could not retrieve API token from Magento Instance (Could not find     
         Magento Backend Instance at MAGENTO_BACKEND_BASE_URL or                
         MAGENTO_BASE_URL)                                                      
         Please verify if MAGENTO_BASE_URL, MAGENTO_ADMIN_USERNAME and          
         MAGENTO_ADMIN_PASSWORD in .env are valid.                              

 Connecting to Selenium Server ...

 [OK] Successful                                                                

 Loading Admin page ...

 [OK] Successful                                                                

 Loading Storefront page ...

 [OK] Successful                                                                

 Running Magento CLI ...

 [ERROR] Failed to run Magento CLI command                                      
         Please reference Magento DevDoc to setup command.php and .htaccess     
         files.
```

So, it appears it is correctly connecting to Selenium (although failing on the Magento CLI command). The "could not retrieve token" appears to be coming from some place else.

I can look at this deeper but will need a bit more time.

dmanners:
Nice, the command stuff needs a change in the nginx.conf I think I commented about this in the READM.md changes. I was not sure how best to automate this change though.

eduardo-g-silva:
Hi have you manage to make mftf to build and run ?

markshust:
I completely forgot about this PR! Thanks @eduardo-g-silva for chiming back in on this. I will take a look at it shortly. As soon as I can confirm it's working, I'll merge this in, or if someone else can confirm it's working I will also do this, since this PR doesn't seem to break anything.

Issue #134, opened 2019-03-09
Load db credentials from env file during setup


shochdoerfer:
Currently the db configuration is stored in two locations, in the env file for docker / docker-compose and in bin/setup.sh. It would be nice to have the bash script load the environment settings from the env file as well.

For Bash you could use something similar to this - that works fine in our setup:
```
DIR="${BASH_SOURCE%/*}"
if [[ ! -d "$DIR" ]]; then DIR="$PWD"; fi
source "$DIR/../env/db.env"
```

For the Windows scripts I have no idea ;(

markshust:
sorry didn't see this issue until just now. i don't like the env vars leaking into the global namespace... however, perhaps we can import the db.env file into `bin/setup`, and reference `bin/magento setup:install` command to look at those env vars instead of the hardcoded vals?

shochdoerfer:
no worries.

Sure that's possible That's exactly what my code snippet does. See [here](https://github.com/bitExpert/magento2-force-login/blob/master/bin/setup#L5) how I did it in one of our modules. The only thing I do not know is, if a similar approach is available on Windows. I guess we would need a solution that works on both platforms.

PR #302, opened 2020-08-31
load db credentials from db.env in setup script similar to mysql script

piotrkwiecinski:
closes: #134


markshust:
Awesome, thanks @piotrkwiecinski! You've been a rock star lately 🚀

Issue #244, opened 2020-03-21
Implement alpine/micro containers for all images, prod support


markshust:
The current images are quite large, and could potentially take up a couple of gigs of space. This probably isn't a big deal for local dev, however it makes sense to look into more efficient ways of creating these images so they can also be used on production.

We'd need a way to build base images which could be used in both dev & prod, and also the ability to override the images for dev-specific packages. For example, even though `vi` is a very small package (and this may be a construed example), it's fairly common practice to not include this package on prod as it's just really not needed, however it is needed a lot in dev. We'd also probably want things like xdebug installed & available in the dev images, but not required or needed on prod, so we should save the space on prod and not include it.

shochdoerfer:
We extended the `bin/setup` script to install linux tools for local development but not have them in the images built for production. Was the best option than having to deal with multiple images and keep them in sync.

PR #306, opened 2020-08-31
nginx alpine image

piotrkwiecinski:
@markshust I'm not sure if this should be just an override or should I create a separate dir something like `1.18-alpine` and put a docker file there.

Refs: #244


markshust (approved):

markshust:
Awesomesauce!!! Thanks so much for this (and sorry it took so long for me to get around to it). I tested this out on Magento 2.3.4 as well as 2.4.0 and everything works great. Merging in and will be building shortly.

PR #301, opened 2020-08-29
Automated tests with bats

piotrkwiecinski:
This is a teaser up for a discussion. Automated tests for scripts located in bin directory.

It `bats` in the path. Link to [bats installation](https://github.com/bats-core/bats-core/blob/v1.2.0/README.md#installation)

Tested with v1.2.0.

Installation of bats on MacOs:

```
brew install bats-core
```

From the repo root:

```
bats compose/tests
```

I've added a dump check if the docker is setup. If not it downloads m 2.4.0 and runs `bin/setup` so first run takes ages.

PR #307, opened 2020-09-02
ensure scripts can be run outside docker compose root dir, fail on error

piotrkwiecinski:
The implementation enables running scripts from other than docker compose root dir.

Example from  `src`:

```
../bin/start
```

Example from parent dir:

```
compose/bin/start
```

I've tried to follow some best practices + apply `shellcheck` recommendations.

Source files checked with `shellcheck` version `0.7.0`:
```
shellcheck -x -P compose/bin -P compose/env compose/bin/*
```

Some reference:

- https://google.github.io/styleguide/shellguide.html
- https://bash3boilerplate.sh/

Is should help simplify: #301


compose/bin/clinotty
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
-[ -z "$1" ] && echo "Please specify a CLI command (ex. ls)" && exit
-docker-compose exec -T phpfpm "$@"
+[ $# -eq 0 ] && echo "Usage: ${0} command" && exit
```
piotrkwiecinski, line 2: I'm not sure if the usage is a better version. It can be reverted.

piotrkwiecinski:
@markshust I belive this one is ready to test. Except of stopping some commands on the first error there should be no functional changes.

markshust:
I'm going to close this PR because it's just too invasive to the project -- touching too many files and a lot of duplicated code. Perhaps there is a more elegant way to do this.

PR #310, opened 2020-09-05
fix start script on ubuntu 18.04

### LINKED ISSUE #309: Failed to start docker for missing volume files on bin/start (Ubuntu/Linux)
**Description**

OS: 18.04.5 LTS 
Docker version 19.03.12
docker-compose version 1.24.0

When trying to start Docker containers with bin/start, the Docker containers don't start because of "missing volume files".
The files are there, and starting containers with `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d --remove-orphans` works fine.

Looks like that the additional check for existing folders in bin/start "Check if volume files exist to avoid creating an empty folder" causes the issue. The parseYaml function seems to remove a character in the volume paths: see below console output **./src/app/et:** instead of **./src/app/etc/** .

Console output

```./src/app/cod: No such file or directory
./src/app/desig: No such file or directory
./src/app/et: No such file or directory
./src/composer.jso: No such file or directory
./src/composer.loc: No such file or directory
./src/nginx.conf.sampl: No such file or directory
Failed to start docker for missing volume files
```

**Steps To Reproduce**
1. cd into the ~/Sites/magento2 folder
2. run bin/start

**Expected Result**
All Docker containers are started

**Actual Result**
Docker container don't start. Instead getting the error message "Failed to start docker for missing volume files"

piotrkwiecinski:
fixes #309

```
substr(string, start [, length ])
If start is less than one, substr() treats it as if it was one. (POSIX doesn’t specify what to do in this case: BWK awk acts this way, and therefore gawk does too.) If start is greater than the number of characters in the string, substr() returns the null string. Similarly, if length is present but less than or equal to zero, the null string is returned.
```

Apparently Ubuntu 18.04 had an older version of awk which didn't handle above the same way.

### DIFF
--- a/compose/bin/start
+++ b/compose/bin/start
@@ -20,7 +20,7 @@ function parseYaml {
     if (length($3) > 0) {
       vn=""; for (i=0; i<indent; i++) {vn = (vn)(vname[i])("_")}
       if ("'$2'_" == vn) {
-         print substr($3 ,0 , match($3,":")-1)
+         print substr($3 ,1 , match($3,":")-1)
       }
     }
   }'


piotrkwiecinski:
@ennostuurman would you be able to test the fix above? It solved the problem for me on virtual machine with ubuntu 18.04.

markshust (approved):

ennostuurman:
@piotrkwiecinski yes this fix works on Ubuntu 18.04. Thanks!

cc @markshust

UrielHernandezR:
I have the same problem with `bin/start` on Mac Os M2 ventura.

PR #326, opened 2020-10-17
Check Elasticsearch connection before setup:install

requix:
There is no sense to continue installation process in case when Elasticsearch service isn't available. 
We will try to connect every 5 sec during next 30 sec. And script is stopped if it's unsuccessful.

This helps to prevent failed Magento installation in next cases:
1. Elasticsearch container is running, however Elasticsearch service isn't ready
2. Elasticsearch container was killed because of RAM issues.

So we can detect Elasticsearch issues on early stage.

Otherwise, we will see an error during installation:
`Could not validate a connection to Elasticsearch. No alive nodes found in your cluster`

### DIFF
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -2,6 +2,9 @@
 source env/db.env
 BASE_URL=${1:-magento2.test}
 
+ES_HOST=elasticsearch
+ES_PORT=9200
+
 bin/stop
 
 docker-compose -f docker-compose.yml up -d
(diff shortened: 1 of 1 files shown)


rangerz:
`until` statement seems unable printf dot(...) for processing...

```
until sleep 10; do
    printf '.'
    sleep 1
done
```

requix:
@rangerz, in your example commands inside loop will be never executed. If you would like to test `until` statement, try this code: `timeout 10 bash -c "until false; do printf '.'; sleep 1; done"`

markshust (approved):

markshust:
Thanks a bunch, merged in. I'll do some testing before tagging new version 35.0.0 but looks good.

Issue #327, opened 2020-10-17
bootstrap issue - trying to use bin/download, bin/setup with 2.4.1 - enterprise


carylewis:
When attempting to use the onelinesetup script to install magento 2.4.1 from the magneto repos, the script fails.

**Steps To Reproduce**
1. curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/onelinesetup | bash -s -- magento2.test 2.4.1

**Expected Result**

magento 2.4.1 is installed

**Actual Result**

 nothing is installed

The issue is that bin/download (as previously reported), has an issue detecting that the 2.4.1.tar.gz file is not available yet.

That is a quick fix: just update all the "1M" with "10M" in bin/download.

But there is a bigger issue:

If you want to install 2.4.1 CE or a EE version, you have to modify bin/download, but then if you try to run it, there aren't any containers yet, so the remote composer execution fails.

I tried to run the bin/download 2.4.1 after installing 2.4.0, but then the bin/setup script doesn't work because it tries to copy all the files from the host to the container, and the download script does a rm -fr src. This results in the copytocontainer copying from the root of the host to the container.

I was able to upgrade from CE to EE 2.4.1.

Using the modified bin/download, I then copied from the container's /var/www/html/src folder back to the host's src folder, and then manually flushing the cache, running bin/magento setup:upgrade

But if someone has a better suggestion that would be appreciated.

piotrkwiecinski:
@carylewis I've created a PR to add a second parameter to the download script to allow creating project from enterprise repo.

Until it's merged what you can do is use the version of a download scipt from the PR.
After you clone current repo what you can do is. 

```
bin/removeall
rm -r src
bin/download 2.4.1 enterprise
bin/setup your-project.domain
```
If you have any issues with the command it the PR let me know. 
I have my composer set up to access EE repository and I was able to set up fresh project using the following commands.

carylewis:
Thank you for the quick reply. Hopefully your PR will be merged quickly.

markshust:
All merged, thanks everyone! I don't think updating 1M to 10M would do anything, as a failed archive should always have a value under 1M. Please let me know if I'm wrong. Cheers -Mark

markshust:
Merged your other PR @piotrkwiecinski -- like the `-f` flag!

PR #329, opened 2020-10-17
support for magento 2 enterprise download

piotrkwiecinski:
Refs #327


markshust (approved):

markshust:
I added `rm -rf src` before the enterprise composer install, as it could fail if the `src` folder exists. Otherwise looks good, thanks a lot!

PR #335, opened 2020-10-22
Add setup-composer-auth

rangerz:
`setup-composer-auth` can setup `~/.composer/auth.json` for using composer in `docker-compose.yml`

```
    volumes: &appvolumes
      - ~/.composer:/var/www/.composer:cached
```

The auth setting can from `~/src/auth.json` or manual input, and can use `composer` or `python3` on host


markshust (approved):

markshust:
Thanks @rangerz !

PR #339, opened 2020-10-28
added --prefer-dist in composer install command

milindsingh:
Added `--prefer-dist` in composer install command to solve:

1. error on installing from source  `connection reset by peer`
2. speed up the dependencies download.


markshust:
Thanks @milindsingh, looks good!

PR #341, opened 2020-10-29
set composer to latest v1

torhoehn:
Instead of updating the composer version flag regularly there is a flag to keep composer v1.


markshust:
I think it's better if we actually keep a pegged version. I've had issues in the past where Composer was updated to a newer version, and it would break something in the dependency chain.

torhoehn:
Okay, I understood. But with the release of v2 there shouldn't be any critical changes anymore.

markshust (approved):

markshust:
Allright, good argument. All merged in, thanks! 👍

torhoehn:
Thank you! Hopefully we have Composer v2 compatibility soon, so we can go back to a fixed version.

PR #347, opened 2020-11-16
mkcert -install is being called in the wrong place

rnicklin:
In bin/setup-ssl 
There is no CA file to read  in order to call bin/setup-ssl-ca. You must create it first.

In bin/setup-ssl-ca
The `mkcert -install` is in the wrong place, the statement will never be reached unless the CA is already there.


markshust (approved):

markshust:
@rnicklin thank you for your contribution, all merged in!

markshust:
FYI I reverted this PR, it caused issues where the SSL cert was not setup properly. You can verify this on a new install.

Issue #351, opened 2020-12-02
sed error on macOS Big Sur (unused label)


r-martins:
**Description**
Warning error after running `bin/start`

**Steps To Reproduce**
1. run `bin/start`  (MacOS Big Sur - I don't know if this is relevant)

**Expected Result**
Start containers with no error

**Actual Result**
sed: 2: "s|^\([[:space:]]*\)\([a ...": unused label '1;s|^\([[:space:]]*\)\([a-zA-Z0-9_]*\)[[:space:]]*:[[:space:]]*\[[[:space:]]*\(.*\)[[:space:]]*,[[:space:]]*\(.*\)[[:space:]]*\]|\1\2: [\3]\n\1  - \4|;t1'
sed: 2: "s|^\([[:space:]]*\)-[[: ...": unused label '1;s|^\([[:space:]]*\)-[[:space:]]*{[[:space:]]*\(.*\)[[:space:]]*,[[:space:]]*\([a-zA-Z0-9_]*\)[[:space:]]*:[[:space:]]*\(.*\)[[:space:]]*}|\1- {\2}\n\1  \3: \4|;t1'

<img width="1872" alt="image" src="https://user-images.githubusercontent.com/191149/100942708-ee33e900-354f-11eb-9b0c-518c77f584db.png">

markshust:
Can someone else confirm this? I will be upgrading my backup computer to Big Sur over the weekend and can test this out. No problems on previous macOS.

zachwreed:
I can confirm. Noticed that this morning after updating to Big Sur.

markshust:
Thanks guys. I still need to upgrade to Big Sur but will shortly :)

smr-deploy:
Replacing sed with homebrew's gnused solves this problem.

markshust:
Arg, ok thanks for the update. Just got Big Sur installed on my backup computer and will look for a fix shortly.

markshust:
So sorry this took so long. This has been a fun year 🙃

I finally managed to install docker-magento on Big Sur. It appears the install completed successfully, it's just that these sed errors show when starting containers. I don't think they cause anything to not work correctly, but it is a nag error.

I'm going to loop in @rangerz as he added the functionality to check to verify if volume mounts exist in the docker-compose.yml file in this commit https://github.com/markshust/docker-magento/commit/12ce5eff69872a29b7ad89b000ae6e7b882a197c

[...]

PR #355, opened 2020-12-15
Fix bin/start for macOS Big Sur

rangerz:
Fixed for #351

Also add refs link and tested on mac and linux.

[How can I parse a YAML file from a Linux shell script?](https://stackoverflow.com/a/51789677/9821321)


piotrkwiecinski:
I can confirm it works on linux.

markshust (approved):

markshust:
@piotrkwiecinski awesome!!! Didn't see this until now. This will be in 35.0.0, to be released momentarily.

PR #371, opened 2021-01-13
Make xdebug command understand partials

r-martins:
This is for lazy ones.. 
Now you can use `bin/xdebug en`, `bin/xdebug dis` or any other thing that starts with `d` or `e` to disable or enable xdebug. :O


markshust (approved):

markshust:
Ha... I like this. All merged in 👍

PR #331, opened 2020-10-19
Improve bin/start and remove backslashes

rangerz:
### Description
1. Use `$(...)` (new style) to replace<code>`</code> backslashes (old style) because this [link](https://unix.stackexchange.com/questions/5778/whats-the-difference-between-stuff-and-stuff/5782#5782)

2. Improve `bin/start` to wait for all services of containers to be ready. 

No need the sleep any more `sleep 5 #Ensure containers are started...`

The concept is the same with #326 (Check Elasticsearch connection before setup:install)

I think the checker put to `bin/start` will be more fit, and also extend more services, such as nginx, mysql, redis, and rabbitmq

The checker also scans service ports to be available before docker start, and scan allocated ports by services after docker start.

If service ports are already allocated, `bin/start` will appear below error message.

<details>
  <summary>bin/start error for allocated port</summary>

```
bin/start
Starting a4f172b22c65_magento2test_rabbitmq_1 ...
Starting magento2test_elasticsearch_1         ...
Starting a4f172b22c65_magento2test_rabbitmq_1 ... error
Starting magento2test_redis_1                 ...
Starting magento2test_db_1                    ...
WARNING: Host is already in use by another container
WARNING: Host is already in use by another container

ERROR: for a4f172b22c65_magento2test_rabbitmq_1  Cannot start service rabbitmq: driver failed programming external connectivity on endpoint a4f172b22c65_magento2test_rabbitmq_1 (4a68cb572cc5641cea47c9292b1d0ce2edff8816f71f3a71e01e03884ad3a86c): Bind for 0.0.0.0:15672 failed: port is alStarting magento2test_db_1                    ... error

ERROR: for magento2test_db_1  Cannot start service db: driver failed programming external connectivity on endpoint magento2test_db_Starting magento2test_mailhog_1               ... error
WARNING: Host is already in use by another container
Starting magento2test_elasticsearch_1         ... error
ERROR: for magento2test_mailhog_1  Cannot start service mailhog: driver failed programming external connectivity on endpoint magento2test_mailhog_1 (71980bcf3e48e9b9fd8e21c433dfe0e70ae5418ff7bbc1eae54d386f555f0afb): Bind for 0.0.0.0:8025 failed: port is already allocated
WARNING: Host is already in use by another container

Starting magento2test_redis_1                 ... done
[...]

PR #384, opened 2021-02-01
Add docker-compose.healthcheck.yml

rangerz:
According to the `HEALTHCHECK` for docker containers I mentioned in [here](https://github.com/markshust/docker-magento/pull/331#issuecomment-770102708)

@markshust, please rebuild all of php (7.3, 7.4, 8.0) and elasticsearch(6.8, 7.6) images. Thanks  :+1:

In `bin/status` also can view all health status
```
magento2.test git:(master) ✗ ./bin/status                                                        git:(master|●4✚2…5
            Name                          Command                  State                         Ports
------------------------------------------------------------------------------------------------------------------------
magento2test_app_1             /docker-entrypoint.sh ngin ...   Up (healthy)   80/tcp, 0.0.0.0:80->8000/tcp,
                                                                               0.0.0.0:443->8443/tcp
magento2test_db_1              /docker-entrypoint.sh --ma ...   Up (healthy)   0.0.0.0:3306->3306/tcp
magento2test_elasticsearch_1   /usr/local/bin/docker-entr ...   Up (healthy)   0.0.0.0:9200->9200/tcp,
                                                                               0.0.0.0:9300->9300/tcp
magento2test_mailhog_1         MailHog                          Up (healthy)   0.0.0.0:55038->1025/tcp,
                                                                               0.0.0.0:8025->8025/tcp
magento2test_phpfpm_1          docker-php-entrypoint php-fpm    Up (healthy)   9000/tcp, 9001/tcp
magento2test_rabbitmq_1        docker-entrypoint.sh rabbi ...   Up (healthy)   15671/tcp, 0.0.0.0:15672->15672/tcp,
                                                                               25672/tcp, 4369/tcp, 5671/tcp,
                                                                               0.0.0.0:5672->5672/tcp
magento2test_redis_1           docker-entrypoint.sh redis ...   Up (healthy)   6379/tcp
```

Refs: 
[magento/magento-cloud-docker](https://github.com/magento/magento-cloud-docker/tree/develop/images)
[sun-asterisk-research/docker-php](https://github.com/sun-asterisk-research/docker-php/tree/master/7.4/buster/fpm)


compose/bin/setup
```diff
@@ -64,8 +52,8 @@ bin/clinotty bin/magento setup:install \
   --session-save-redis-log-level=4 \
   --session-save-redis-db=2 \
   --search-engine=elasticsearch7 \
-  --elasticsearch-host=$ES_HOST \
-  --elasticsearch-port=$ES_PORT \
+  --elasticsearch-host=elasticsearch \
+  --elasticsearch-port=9200 \
```
markshust, line 56: We'll want to back out this update so a user can override the Elasticsearch host/port with env vars.

compose/bin/setup
```diff
@@ -3,14 +3,10 @@ set -o errexit
 source env/db.env
 BASE_URL=${1:-magento2.test}
 
-ES_HOST=elasticsearch
-ES_PORT=9200
```
markshust, line 7: Need to add these back in

compose/docker-compose.yml
```diff
@@ -19,12 +19,27 @@ services:
       - appdata:/var/www/html
       - sockdata:/sock
       - ssldata:/etc/nginx/certs
+    healthcheck:
+      test: 'curl --fail http://127.0.0.1:8000'
+      interval: 1m30s
```
markshust, line 24: Why is the `interval` a different value from the `docker-compose.dev.yml` file?

compose/docker-compose.yml
```diff
@@ -19,12 +19,27 @@ services:
...
+      timeout: 10s
```
markshust, line 25: Same with timeout, I would think these would be consistent between the files unless there was some other reason to do so (which should probably then be documented).

compose/docker-compose.yml
```diff
@@ -53,6 +78,11 @@ services:
       - "5672:5672"
     volumes:
       - rabbitmqdata:/var/lib/rabbitmq
+    healthcheck:
+      test: 'rabbitmq-diagnostics -q ping'
+      interval: 30s
```
markshust, line 83: The healthcheck params are repeated a few times, can we make this into a YAML anchor like volumes are done?

markshust (changes requested):
Very cool, thanks for this. Some questions/comments posted.

compose/bin/setup
rangerz: ok, I will recover this part.

But I think this setting is very common. 
If someone need change Elasticsearch host/port, change here is not enough.

docker-compose.yml (just guess, not test yet)
```
elasticsearch:
    image: markoshust/magento-elasticsearch:7.6.2-2
    ports:
      - "9200:9200" # this also need to change
      - "9300:9300"
    environment:
      - "discovery.type=single-node"
      - ES_HOST=elasticsearch # add new value
      - ES_PORT=9200               # add new value
```

compose/docker-compose.yml
rangerz: I think remove all interval, timeout, and start-period will be fine.
Using their default value may a better choice

The options that can appear before CMD are:

```
--interval=DURATION (default: 30s)
--timeout=DURATION (default: 30s)
--start-period=DURATION (default: 0s)
--retries=N (default: 3)
```

compose/docker-compose.yml
rangerz: Remove all interval, timeout, and start-period, and use default value.

compose/docker-compose.yml
rangerz: Remove all interval, timeout, and start-period, and use default value.

compose/docker-compose.yml
markshust: Cool, definitely prefer the defaults 👍

compose/docker-compose.yml
markshust: 👍

compose/docker-compose.yml
markshust: 👍

compose/bin/setup
markshust: I think this is where we need updates. These settings are file, but need to pull in from envvar if it exists.

I didn't test but something like this pseudo code may work:

```
ES_HOST="${ES_HOST:=elasticsearch}"
ES_PORT="${ES_PORT:=9200}"
```

This way we don't need to define anything in the XML files.

compose/bin/setup
markshust: Yep, I'd avoid the vars being set here though and set defaults within the bash file (see https://github.com/markshust/docker-magento/pull/384#discussion_r570235980). But, this is how & where someone would override these vars. This would be useful to add to the documentation.

compose/bin/setup
markshust: On second thought, maybe be better to create an `env/elasticsearch.env` file for these similar to https://github.com/markshust/docker-magento/blob/master/compose/env/db.env

rangerz: @markshust I think this push request (commit) should only keep changes about `HEALTHCHECK`.
For other environmental variables, I will create another one for it.

compose/bin/setup
rangerz: I think this push request (commit) should only keep changes about `HEALTHCHECK`.
For other environmental variables, I will create another one for it.

rangerz:
@markshust 

I added `docker-compose.healthcheck.yml` for `bin/start` without until statement in bash script

Please review it.

markshust (approved):

markshust:
Hi @rangerz, all looks good! But I don't see any additions to `bin/start`. Did you want me to merge this for testing, or is this missing an update in `bin/start`?

I just merged your other big PR in for docker-compose/docker compose, so perhaps that needed to be merged in first?

rangerz:
@markshust Yes, here is waiting for `bin/docker-compose` commit.

It's updated to put `docker-compose.healthcheck.yml` into `bin/docker-compose` and then remove `until` wait for RabbitMQ and Elasticsearch services

markshust (approved):

markshust:
Awesome... this is sweet!!! 🧡🧡🧡

Issue #385, opened 2021-02-02
`fixowns` fails with "Operation not permitted"


geoffjukes:
On Big Sur, in `/Users/geoff/magento`, execute:
```
curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/onelinesetup | bash -s -- magento2.test with-samples-2.4.1
```
the `fixowns` process fails with a screen full of:

```
chown: changing ownership of ‘/var/www/.composer/cache/files/wikimedia/less.php’: Operation not permitted
```
Then the next processes fail with similar issues.

markshust:
Comment left in PR.

markshust:
Closing per https://github.com/markshust/docker-magento/pull/386#issuecomment-772932552

PR #386, opened 2021-02-02
fix operation not permitted

piotrkwiecinski:
Possible fix for #385

@markshust @geoffjukes would you be able to test it?

### DIFF
--- a/compose/bin/fixowns
+++ b/compose/bin/fixowns
@@ -2,7 +2,7 @@
 echo "Correcting filesystem ownerships..."
 
 if [ -z "$1" ]; then
-  bin/rootnotty chown -R app:app /var/www/
+  bin/rootnotty chown -R app:app /var/www/html
 else
   bin/rootnotty chown -R app:app /var/www/html/"$1"
 fi


geoffjukes:
@piotrkwiecinski Progress....

Now the `fixowns` passes, but then fails with:

```
Correcting filesystem permissions...
find: ‘vendor’: No such file or directory
Filesystem permissions corrected.
Global composer auth already exists
Forcing reinstall of composer deps to ensure perms & reqs...
Changed current directory to /var/www/.composer
 [ErrorException]                                                             
 fopen(/var/www/.composer/cache/repo/http---repo.packagist.org/p-provider-2020-04.json): failed to open stream: Operation not permitted
```

markshust:
@geoffjukes is your `~/.composer` or home `~` directory listed under Docker Preferences > Resources > File Sharing? Usually this is taken care of with `/Users`. Almost every error I've seen like this has stemmed from someone changing or removing the `/Users` line, as there should never be any permissions issues with this project (as long as you also don't use `bin/root` deviously).

Screen Shot 2021-02-03 at 9 18 32 AM

geoffjukes:
@markshust `/Users` is showing under Resources. I'll keep poking around, see if something is wacky with Docker

markshust:
Ok glad it's all resolved! I'll close out this PR as we don't likely need it then.

PR #387, opened 2021-02-02
fix shellcheck warnings

piotrkwiecinski:

PR #388, opened 2021-02-02
shellcheck fix with github action

piotrkwiecinski:
https://github.com/markshust/docker-magento/pull/387 but with Github Action

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,24 @@
+name: 'ShellCheck'
+
+on: 
+  push:
+    branches: ["master"]
+  pull_request:
+
+jobs:
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout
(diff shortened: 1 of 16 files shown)


piotrkwiecinski:
@markshust here is an example how it works: https://github.com/piotrkwiecinski/docker-magento/pull/1

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Run ShellCheck
+      uses: ludeeus/action-shellcheck@master
```
piotrkwiecinski, line 18: There is also a tagged release if you think it would be safer

markshust: I actually think `master` here is probably just fine, if there is ever an issue or something weird going on, we can disable the action or look at it at that time.

markshust (approved):

PR #399, opened 2021-02-05
Added Makefile with list of available commands.

peakercope:
Hello!
I created Makefile with all CLI commands.
I think it can be helpful to see this list of commands with description right into your terminal instead of opening github page.
To see all commands you can type `make help` or just `make`.

### DIFF
--- a/compose/Makefile
+++ b/compose/Makefile
@@ -1,0 +1,178 @@
+SHELL := /bin/bash
+
+args = `arg="$(filter-out $(firstword $(MAKECMDGOALS)),$(MAKECMDGOALS))" && echo $${arg:-${1}}`
+
+green  = $(shell echo -e '\x1b[32;01m$1\x1b[0m')
+yellow = $(shell echo -e '\x1b[33;01m$1\x1b[0m')
+red    = $(shell echo -e '\x1b[33;31m$1\x1b[0m')
+
+format = $(shell printf "%s %-38s %s" "$(call yellow,$1)" "$(call green,$2)" $3)
+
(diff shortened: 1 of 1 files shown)


markshust:
I love this idea!

I see a bunch of gobbly gook when trying to run `make`:
Screen Shot 2021-02-10 at 9 13 30 AM

I have a feeling this should be something like a `bin/help` script instead.

peakercope:
@markshust yes, it is
seems it some platform incompatibility
I tested in on Linux. Which platform do you use?

peakercope:
by the way here is example of my output
image

markshust:
@peakercope I'm on macOS 11.2. We will definitely want to support macOS, as I try to make sure everything is compatible on macOS/Linux/etc.

peakercope:
@markshust  you right. it was an issue with `echo -e` command on macOS system.
I've fixed it, now the output should look like this
image

drpayyne:
Hi @markshust, could you please have a look at this one? This looks ready and would be very useful 😄

markshust:
@drpayyne thanks for the ping, I missed the last update from @peakercope. My main computer is down for another week or so, but I will look at this as soon as I get my more powerful computer back. Working on a MacBook Air is a bit painful at the moment 😅

peakercope:
Hello @markshust !
Thanks for merging my PR

Please take a look at [https://github.com/markshust/docker-magento/pull/547](https://github.com/markshust/docker-magento/pull/547) I've added some missed commands.

PR #401, opened 2021-02-07
introduce magento.env to customize install parameters

### LINKED ISSUE #312: Introduce magento.env file to allow overriding magento install parameters
Currently admin user, admin front name, currency and locale are hardcoded in the `bin/setup` script.

Add support to load above values from `env/magento.env` file with fallback to previously hardcoded defaults.

image

piotrkwiecinski:
closes #312


markshust (approved):

markshust:
Very nice!

Issue #405, opened 2021-02-08
docker-compose.dev.yml Linux setup causes 403


mattgrul:
**Description**
Modify the docker-compose.dev.yml file for Linux use if the nginx.conf file is commented out as well this results in a 403 Forbidden error.

**Steps To Reproduce**
1. Modify docker-compose.dev.yml to the below:
```
      # Host mounts with performance penalty, only put what is necessary here
      #- ./src/app/code:/var/www/html/app/code:cached
      #- ./src/app/design:/var/www/html/app/design:cached
      #- ./src/app/etc:/var/www/html/app/etc:cached
      #- ./src/composer.json:/var/www/html/composer.json:cached
      #- ./src/composer.lock:/var/www/html/composer.lock:cached
      #- ./src/nginx.conf.sample:/var/www/html/nginx.conf:cached
      #- ./src/auth.json:/var/www/html/auth.json:cached
      #- ./src/m2-hotfixes:/var/www/html/m2-hotfixes:cached
      #- ./src/patches:/var/www/html/patches:cached
      #- ./src/var/log:/var/www/html/var/log:cached
      #- ./src/var/report:/var/www/html/var/report:cached
      # To sync your SSH to the container, uncomment the following line:
      #- ~/.ssh/id_rsa:/var/www/.ssh/id_rsa:cached
      # Linux only: remove the above lines and mount the entire src directory with:
      - ./src:/var/www/html:cached
```
2. Load https://magento2.test/ - 403 Forbidden 

**Expected Result**
The domain should load with no issues.

**Actual Result**
The domain loads with a 403 error

vy-shmal:
Uncomment this line **- ./src/nginx.conf.sample:/var/www/html/nginx.conf:cached**  as you can see it changes the name of the file from **nginx.conf.sample** to **nginx.conf** . Or you can also not uncomment the line but **rename** the file in your root folder from **nginx.conf.sample** to **nginx.conf.** 

```# Host mounts with performance penalty, only put what is necessary here
      #- ./src/app/code:/var/www/html/app/code:cached
      #- ./src/app/design:/var/www/html/app/design:cached
      #- ./src/app/etc:/var/www/html/app/etc:cached
      #- ./src/composer.json:/var/www/html/composer.json:cached
      #- ./src/composer.lock:/var/www/html/composer.lock:cached
      - ./src/nginx.conf.sample:/var/www/html/nginx.conf:cached
[...]

PR #411, opened 2021-02-18
Change in note for linux users

r-martins:
As mentioned by @vy-shmal [here](https://github.com/markshust/docker-magento/issues/405#issuecomment-776029611), if we comment nginx.conf line, we don't get Magento's nginx configuration, causing 403 error.


markshust (approved):

markshust:
Ahhh thanks @r-martins! I never noticed that.

PR #441, opened 2021-04-01
add varnish container with configurations

epolish:


st3phani (approved):
Very good

epolish:
There is a problem with ssl termination...
Varnish can't handle ssl traffic itself... so we need something like nginx to proxy the traffic.
But I'll research some options with varnish official ssl offloader - hitch

chrom:
add varnish container  - good Idia!
ability to run many Magento instances at once - it already looks like kubernetes is suitable here

st3phani (approved):

mucan54:
Anyone needs varnish can use this command for online setup:

curl -s https://raw.githubusercontent.com/mucan54/docker-magento-varnish/master/lib/onelinesetup | bash -s -- magento.test 2.4.6 community

markshust:
Wow, it's been a while. I'm sorry for the delay — too many issues/PRs were lingering for a while and I just had a bot clean them all up.

I think I will take this in, but will need to create this custom Varnish image and push it up to Docker Hub, so it's consistent with the other Docker images.

I do not want to make this default, but opt-in. This means that we need to add some documentation to the README that informs everyone how to use it. The varnish-related stuff in compose.yaml and compose.dev.yaml should be commented out by default. The Nginx config should be put into a new `default.varnish.conf` file, and again, instructions can inform users how to use this file (needing to define custom Nginx file, make changes to compose.yaml, etc.).

I think the vcl's should also be volume-mounted so they can be easily changed?

markshust:
Thanks for kicking this off back in 2021, @epolish — Varnish support is still something I'd like to land, but this branch can't be merged as-is. The project has moved on enough that a rebase would leave essentially nothing of the original diff intact:

- Compose files moved from `compose/docker-compose.yml` to root-level `compose.yaml` / `compose.dev.yaml`
- Nginx 1.18 is no longer the default — supported versions are now 1.22 / 1.24 / 1.26 / 1.28
- Varnish 4.0 is end-of-life; Magento 2.4 supports 6.x / 7.x
- GitHub flags this as conflicting

If anyone (original author or otherwise) wants to take another swing at this, here's the shape I'd merge — keeping it opt-in and non-invasive so the 95% of users who don't need Varnish aren't affected:

- A `markoshust/magento-varnish` image published to Docker Hub (consistent with the other images here), Varnish 7.x
- Varnish service entries in `compose.yaml` / `compose.dev.yaml` **commented out by default**
- A separate `default.varnish.conf` for nginx (don't modify the existing `default.conf`) — users opt in by swapping which conf gets mounted
- VCL files volume-mounted so they're easily customizable
- README section documenting how to enable it

Going to close this one to keep the queue tidy, but a fresh PR along those lines is very welcome.

PR #454, opened 2021-05-01
Updated text in fixowns & fixperms

drpayyne:
Updated text in the above bash files to reflect the command name. `fixowns` will "fix" ownerships instead of "correcting" them. Similarly with `fixperms`


drpayyne:
Hi @markshust, please review when you can. Thanks!

markshust (approved):

markshust:
Thank you, all merged

PR #465, opened 2021-05-28
Added description how to  disable multifactor auth

Mozart409:


Mozart409:
I think adding this to the main readme would be a better option. I did not find it and I opened an issue.

markshust:
I'm closing this out, this is not specific to this repo, but in Magento in general. No reason to document it here.

PR #513, opened 2021-08-19
Added Elasticsearch support for ARM environments

drpayyne:
#### ADDED

- Elasticsearch Docker image for `v7.9.3`
- A new Compose file with custom/necessary configuration for ARM environments. This PR adds content to this file to force `v7.9.x` for ARM environments.

### Notes

- Elasticsearch `7.9.x` needs a minimum Magento version of `~2.3.7` or `~2.4.2` for official Magento tested support (https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements.html)
- The new image needs to be built and published to Docker Hub
- Need to figure out on how to possibly automate the inclusion of the ARM Compose file if the `docker-compose up` command is run on an ARM environment.
- The ARM Compose file can be reused to send environment variables for other images (upcoming PRs). For example, we need to switch the `mkcert` and `ioncube` package architectures inside the Dockerfile for cross-compatibility

### DIFF
--- a/compose/docker-compose.arm.yml
+++ b/compose/docker-compose.arm.yml
@@ -1,0 +1,12 @@
+# Mark Shust's Docker Configuration for Magento
+# (https://github.com/markshust/docker-magento)
+#
+# Include this configuration file for ARM environments.
+#
+# Version 38.0.0
+
+version: "3"
+
+services:
+  elasticsearch:
+    image: markoshust/magento-elasticsearch:7.9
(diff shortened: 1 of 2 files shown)


drpayyne:
Note: If no other ARM Support PR needs a new ARM Compose file, we can eliminate the file even for this and add a note in the documentation saying something like "A minimum of Elasticsearch 7.9 is required for ARM environments". For more info, please refer note 3 in https://github.com/markshust/docker-magento/pull/515.

compose/docker-compose.arm.yml
```diff
@@ -0,0 +1,12 @@
...
+#
+# Include this configuration file for ARM environments.
+#
+# Version 38.0.0
+
+version: "3"
+
+services:
+  elasticsearch:
+    image: markoshust/magento-elasticsearch:7.9
```
drpayyne, line 12: This file isn't _really_ necessary, but is present to demonstrate that ARM machines _need_ `7.9` to get a native image. Versions `<7.8` do not provide multi-platform images.

PR #512, opened 2021-08-19
Replaced mailhog/mailhog with sj26/mailcatcher

drpayyne:
Credit to @luancschmitz for these changes! (added him as co-author)

PR #516, opened 2021-08-19
[ARM Support] Added multi-arch support for PHP

drpayyne:
#### ADDED

- Multi-arch support for the PHP image

### Notes

An environment variable is defined in the `docker-compose.arm.yml` file for specifying the ARM based version. If this file is not parsed, the default value from the `Dockerfile` is read.


drpayyne:
Update: current PR logic will not work. Working on a fix.

drpayyne:
@markshust, I've removed the MailHog binaries from the PHP images in this commit: https://github.com/markshust/docker-magento/pull/516/commits/6cb512018099f5c0a363e46460ab9e9e9858f0be since we're moving from MailHog to a _maintained_ fork of MailCatcher in https://github.com/markshust/docker-magento/pull/512. This needs to be tested.

images/php/7.3/Dockerfile
```diff
@@ -46,14 +46,17 @@ RUN docker-php-ext-install \
   xsl \
   zip
 
+ARG TARGETARCH
+
 RUN cd /tmp \
-  && curl -O https://downloads.ioncube.com/loader_downloads/ioncube_loaders_lin_x86-64.tar.gz \
-  && tar zxvf ioncube_loaders_lin_x86-64.tar.gz \
+  && ARCH=$(if [ "$TARGETARCH" = "arm64" ]; then echo aarch64; else echo x86-64; fi;) \
+  && curl -O https://downloads.ioncube.com/loader_downloads/ioncube_loaders_lin_$ARCH.tar.gz \
```
drpayyne, line 53: The first line helps us make this single image cross-platform. The `TARGETARCH` build argument is auto-set as `arm64` or `amd64` based on the build, but we need the architecture flags `aarch64` and `x86-64` for the second line, hence my logic.

drpayyne:
General note for all our Dockerfiles - we need to rebuild these as cross-platform. I did this locally using `docker buildx build --platform linux/arm64,linux/amd64 .`, but also using the GitHub Action present in my multi-arch fork of @davidalger's PHP repo at https://github.com/drpayyne/docker-images-php-multi-arch/blob/master/.github/workflows/cli-loaders.yml#L25-L31.

Affected PRs
- this
- https://github.com/markshust/docker-magento/pull/513
- https://github.com/markshust/docker-magento/pull/515

PR #520, opened 2021-08-21
Make uid & gid of app user configurable

shochdoerfer:
This makes the user id & group id of the app user configurable during docker build.

Some of our developers have a different user id & group id on the host than the default of 1000 used in the build. This causes permission problems when mounting volumes from the host in the container. This PR fixes this by making the user id & group id configurable during the build process.

The downside of this: Since this is a build setting, the users have to build their own image and cannot use the default one provided by you.

When you are building the images nothing changes as the user id & group id of 1000 is used as a fallback when no id is provided during docker build.

If you want to change the id during build you need to execute the following command:
```
docker build --build-arg COMPOSE_USER=1234 .
```

Also, docker-compose seems to support build args:
```
docker-compose build --build-arg COMPOSE_USER=1234
```

Not sure if this PR makes sense but at least it gives users an option when running into permission problems due to mismatching id's on the host and in the container.

### DIFF
--- a/images/php/7.3/Dockerfile
+++ b/images/php/7.3/Dockerfile
@@ -1,6 +1,8 @@
 FROM php:7.3-fpm-buster
 MAINTAINER Mark Shust <mark@shust.com>
 
+ARG COMPOSE_USER=1000
+
 RUN apt-get update && apt-get install -y \
   cron \
   git \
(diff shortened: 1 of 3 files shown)


markshust:
@shochdoerfer thanks so much for this! Since this is backwards-compatible, I have no issue merging this in.

My only thoughts -- the "compose" in the name `COMPOSE_USER` doesn't make much sense to me -- perhaps you can change this to `APP_ID` since this is to change the id's for the `app` user?

shochdoerfer:
@markshust done as you wish :)

PR #571, opened 2021-11-05
updated ssh alg from RSA keys to Ed25519

mkadsv:
Github is dropping RSA keys and enforcing ED25519.
See [Github SSH Key Docs](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account)

Any users that just created an SSH key (Nov 3, 2021 or later)will run into a permissions error.

> RSA keys (ssh-rsa) with a valid_after before **November 2, 2021** may continue to use any signature algorithm. RSA keys generated after that date must use a SHA-2 signature algorithm. Some older clients may need to be upgraded in order to use SHA-2 signatures.


markshust:
Thanks for the info. Are the new keys default-named to `id_ed25519`? I would think the name `id_rsa` may still be kept for posterity purposes.

If we were to switch this (or add it in), we'd want to still keep the old one there as well so we don't break the setup for existing users.

mkadsv:
If you create new ssh keys in Github the default name is `id_ed25519`. 

To your point about existing users, maybe a simple comment during the [Manual  Setup Guide](https://github.com/markshust/docker-magento#new-projects) for the clone step might be the quickest solution for now:
`# bin/cli git clone git@github.com:magento/magento2.git .`

markshust:
Thank you very much, I confirmed that is indeed the name of the SSH key.

Let me test to see if things still work and there are no weird errors if I add both to the configuration. I'd like to test to see if it still works with an older key present and not a new one, and vice versa. If everything works, I think we can have both entries there for the time being until GitHub nixes RSA's (which will set the standard for everyone moving forward to use the new encryption type).

More info at https://stefanbauer.me/articles/update-your-ssh-keys-to-ed25519

I'll keep this open while I do some additional testing.

h3xx:
I've noticed some issues with even adding this mapping without testing first:

If the `~/.ssh/id_rsa` file doesn't exist before running docker-compose, it'll **create a directory** with the path `~/.ssh/id_rsa/`

Ultimately, I believe the startup scripts should **test the file existence** before mapping it as a volume, but they don't.

markshust:
@h3xx I definitely feel ya and agree with your findings. Going to sit on this a while longer (I know it's been a few years... but maybe need a few more 😅). I don't think there is a good way to conditionally map keys like this with Docker volumes.

Also, I have yet to have any issues or needs to actually create a key with this algo? Is this default/required by GitHub? I've been using the same key for many many years, so a bit out of the loop and not sure if this is still even a thing or needed.

markshust:
Thanks again @mkadsv for raising this back in 2021, and @h3xx for the careful analysis in 2023.

Going to close this one out. A few reasons:

1. GitHub's RSA deprecation ended up only affecting new RSA keys signed with SHA-1 — existing RSA keys still authenticate fine, and we haven't seen reports of breakage in the years since.
2. As-is, the swap would regress every existing user with an `id_rsa` key — and per @h3xx's note, Docker will silently create a directory at the missing key path, which is worse than the original behavior.
3. The real improvement here is conditionally mounting whichever key(s) actually exist (or mounting `~/.ssh` read-only). That's a different shape of change and deserves its own issue if anyone wants to pick it up.

For now, users who only have an ed25519 key can either clone over HTTPS or adjust the volume mount locally. Closing — but happy to revisit in a new issue if this starts affecting more people. 🙏

PR #576, opened 2021-11-12
Update README.md

mattgrul:
Small rename to the correct file where change should be made.


markshust:
It actually is the `docker-compose.yml` file... this changed in 41.0.0.

PR #599, opened 2021-12-21
added script to stop all running container

massalinux:
Hi Mark, i added this script to stop all running container, i found it usefull when you switch project and find yourself in the new project directory and cannot start process due to port already used by other running processes.


markshust:
Cool, I like it! We also need to add a line about it to the README, then I'll merge it in.

massalinux:
here you have it :-)

markshust (approved):

PR #612, opened 2022-01-06
Lock base_url in env.php

tschifftner:
I think sensitive data like the base_url should not be editable and therefore belongs into env.php


markshust:
Appreciate the PR, but I think we should keep this unlocked, so you can easily change the value in development. I wouldn't consider a domain name sensitive data.

PR #633, opened 2022-02-07
Update setup with no interaction

BrentRobert:
The setup is ran within the container, so interaction is not possible.
To not block the setup, --no-interaction should be added.

For example, if an existing project with stores in the config.php get's installed with a clean database, Magento will ask if it should create those store views as well.
The no-interaction defaults to yes in this case.


markshust:
Thanks for this @BrentRobert, going to test this out myself and if everything else still works well, I'll merge this in 👍

markshust:
Sorry it took so long for this! Confirmed there is no issues with this line, so going to merge this in 👍

PR #642, opened 2022-02-21
Add Firefox Xdebug helper, and clarify IDE Key

derickr:
The IDE Key should be a unique name for a project, and unless you use a DBGp proxy or Xdebug Cloud, there is no need for the IDE Key to be anything in particular, or match what is set-up in your browser helper extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ Otherwise, this project now automatically sets up Xdebug support with VS Code. I
 
 ### Xdebug & PhpStorm
 
-1.  First, install the [Chrome Xdebug helper](https://chrome.google.com/webstore/detail/xdebug-helper/eadndfjplgieldjbigjakmdgkmoaaaoc). After installed, right click on the Chrome icon for it and go to Options. Under IDE Key, select PhpStorm from the list to set the IDE Key to "PHPSTORM", then click Save.
+1.  First, install the [Chrome Xdebug helper](https://chrome.google.com/webstore/detail/xdebug-helper/eadndfjplgieldjbigjakmdgkmoaaaoc) or [Firefox Xdebug helper](https://addons.mozilla.org/en-GB/firefox/addon/xdebug-helper-for-firefox/). After installed, right click on the Chrome icon for it and go to Options. Under IDE Key, select PhpStorm from the list to set the IDE Key to "PHPSTORM", then click Save.
 
 2.  Next, enable Xdebug debugging in the PHP container by running: `bin/xdebug enable`.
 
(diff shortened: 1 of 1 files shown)


markshust:
Hi @derickr, thanks for the PR! I like the Firefox additions 👍.

I need to confirm things still work with setting a custom IDE key. Related, in the `php.ini` file is a reference to the IDE Key https://github.com/markshust/docker-magento/blob/master/images/php/7.4/conf/php.ini#L9 -- is this also not needed?

I remember having a lot of issues with things not working when this IDE wasn't set. I'll confirm this is still the case, and if so there may be something else going on here.

derickr:
It's not needed on line #9 for PhpStorm or VS Code. idekeys are only useful when using a DBGp proxy, or netbeans or eclipse which are silly and require it to be set to a specific value for no good reason.

markshust:
@derickr thanks, I will remove & test locally to confirm it's all working. If all looks good I'll merge this in 👍

markshust:
@derickr I've been unable to get Magento debugging configured without setting up an IDE key.

After attempting, this is the notice I receive at the bottom of my IDE:

Screen Shot 2022-03-03 at 12 01 25 PM

Initiating a debugging session from the browser causes this thing to pop up, even though "path mappings" have been set up for my project:

Screen Shot 2022-03-03 at 12 03 06 PM

Screen Shot 2022-03-03 at 12 06 57 PM

If I just OK to that dialog, I get this error:

Screen Shot 2022-03-03 at 12 03 42 PM

The only way for me to get debugging working successfully is to check the "Filter debugging connection..." checkbox:

Screen Shot 2022-03-03 at 12 05 15 PM

Which seems to then properly apply the path mappings listed under that server config:

Screen Shot 2022-03-03 at 12 05 26 PM

Perhaps this is a bug with PhpStorm not properly applying the project-specific path mappings to a config-less Xdebug connection, as that is appears to be going on here (at least that's my understanding).

markshust:
Additionally, under the "Start the debugging session" documentation at https://www.jetbrains.com/help/phpstorm/zero-configuration-debugging.html#start-debugging-session

They show a sample incoming connection:

ps_incoming_debugger_connection

This example doesn't align with Magento, where the request URI doesn't exactly match to the path on the server. The project root needs to map to the `/var/www/html` location on the server:

Screen Shot 2022-03-03 at 12 06 57 PM

markshust:
Closing out this PR, as I cannot get Magento configured property with PhpStorm+Xdebug without using the ide key.

PR #651, opened 2022-03-01
Enable WebP support in GD

bsdnomad:
Enable WebP support in GD

For 7.3 and below the line:

```
RUN docker-php-ext-configure \
  gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/

```
should become:

```
RUN docker-php-ext-configure \
  gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ --with-webp-dir=/usr/include/

```


markshust:
Thanks @bsdnomad, looks good to me! Webp should definitely be in there. No need to add it to the PHP 7.3 images, it hit EOL in November so I'm no longer making updates to those images.

PR #675, opened 2022-04-11
Setup healthcheck timeout env variables

freezy-sk:
For those struggling with setup stuck on **Waiting for connection to Elasticsearch...** or **Waiting for connection to RabbitMQ...**, I've added environment variables for healthcheck timeouts.
I've managed to get setup running with `ES_HEALTHCHECK_TIMEOUT=300` and `RABBITMQ_HEALTHCHECK_TIMEOUT=100`
This pull request also brings RabbitMQ healthcheck through curl instead of just waiting few seconds


markshust:
@freezy-sk this is awesome! Just tested this out and it worked great. This not only helps others who still aren't booted after 10 seconds, but also those who prime quickly are able to exit out of the unnecessary sleep 😁

PR #678, opened 2022-04-13
[hotfix] fixed docker-compose.yml phpfpm version 8 typo

greatdex:
There is no manifest for `magento-php:8.1-fpm-0` which fails at `docker-compose up`


greatdex:
Thanks for the info!

PR #697, opened 2022-04-26
Add setup-domain-multistore script

### LINKED ISSUE #696: Automate configuration of multi store projects
Hi everyone,

I've noticed that we don't have scripts that will automate configuring store access for existing multistore projects.
It would be nice to have a script such as `bin/setup-domain-multistore`.
The idea behind such as script is to run it after the database has been imported and composer install has been run.

It should do the following things:
- update docker-compose.yml and add bind mounts to `nginx.conf` and `default.conf` files
  - I'd add these as bind mounts since doing it that way would allow us to edit the files easily from the outside without the need for an image rebuild. Update the file, restart container and you're done.
- prepend a `mapping` directive to `default.conf` file where each store is accessible at `magento-store-code.test` (this is the simplified approach from the [documentation ](https://devdocs.magento.com/guides/v2.4/config-guide/multi-site/ms_nginx.html#ms-nginx-vhosts)
   - store-code should be fetched from the `store` table and converted to kebab case
 - update base URLs for each store
 - set cookie path and cookie domain config values for each store
 - update `/etc/hosts` and add a domain for each store

This could be extremely useful for existing projects that have many stores (10+). It would allow for faster setup which would lead to less time spent setting up the project, which is a win for all team members. Also, if for some reason, your local environment gets corrupted or is deleted, you could just run the script instead of doing all these steps manually.

TerrorSquad:
I've added a number of files that would allow implementing this script. I think this is a bit rough, so please suggest improvements.

The idea is to first set up a project locally by following the README file. Once the project is up you should do the following steps:
1. Run the `bin/setup-domain-multistore` script
2. Uncomment the two new bind mounts in `docker-compose.linux.yml` file
3. Restart the `app` service
4. Run `sudo bin/add-base-urls-to-hosts`

Notes:
- In order to get the multistore setup working, I have included the latest nginx.conf (`config/nginx/nginx.conf.sample`) and I've added the `config/nginx/default.conf` file (contents taken from [dev docs](https://devdocs.magento.com/guides/v2.4/config-guide/multi-site/ms_nginx.html#ms-nginx-vhosts)).
~~- To fetch the data from `store` table I added a small SQL file that defines the JSON_ARRAYAGG function. The reason for that is: in some older mysql/mariadb versions the function is not available.~~
~~- `store` data is later parsed with `jq` in helper scripts (INTRODUCES [jq](https://linuxcommandlibrary.com/man/jq) as a dependency)~~

~~I used `jq` mainly because I wanted a json structure that I could work with in the bash scripts.~~

Resolves #696


TerrorSquad:
Resolves #696

rangerz:
`jq` is working on MacOS, but other OS may require additional installation.

Another general way to execute command in the container, ex: PHP

```
bin/n98-magerun2 dev:console '$a = 1; call cache:flush; ls; exit;'
```

TerrorSquad:
@rangerz I could then do this `bin/n98-magerun2 sys:store:list --format=json | grep -iv warning | jq`. It will simplify things quite a bit a remove the need for SQL code.

rangerz:
@TerrorSquad Good, it's simpler !

But I think `jq` command should be removed.

Be replaced by common bash command
```
# for example
STORE_ID=$(bin/n98-magerun2 sys:store:list --format=csv | tail -n + 2 | cut -d ',' -f1)
STORE_CODE=$(bin/n98-magerun2 sys:store:list --format=csv | tail -n + 2 | cut -d ',' -f2)
```

TerrorSquad:
@rangerz I've refactored the code to be in line with your example.

markshust:
Is it possible for you to update the readme with instructions on how to use this? Perhaps we can also make the config files contain "multistore" in the filenames, so it is apparent these are only intended for multistore setups?

TerrorSquad:
Hi @markshust I'll update it later today or tomorrow. Did you have the time to test it out?

TerrorSquad:
Hi @markshust  I've updated the readme as well as the names of nginx config files. Since I am working on Linux and don't have a mac I wasn't able to test this out on a mac. Hopefully, it will work the same.

markshust:
@TerrorSquad sorry it took so long for me to get back to this. I just ran shellcheck and it appears to be complaining about a few things. Can you take a look?

```
Notice: ./compose/bin/generate-store-setters:17:22: note: See if you can use ${variable//search/replace} instead. [SC2001]
Notice: ./compose/bin/generate-nginx-mapping:6:8: note: Expressions don't expand in single quotes, use double quotes for that. [SC2016]
Notice: ./compose/bin/generate-nginx-mapping:11:25: note: See if you can use ${variable//search/replace} instead. [SC2001]
Notice: ./compose/bin/add-base-urls-to-hosts:13:22: note: See if you can use ${variable//search/replace} instead. [SC2001]
```

TerrorSquad:
@markshust it should be all good now.

markshust:
Appreciate it. I'm currently re-building the PHP images for NodeJS/NPM support, but will circle back around right after to actually test out this PR against those images, then make a subsequent release if all looks good.

markshust:
@TerrorSquad wow it's been a while since I looked at this one.

I don't have time this week due to holidays and etc., but I'll try to circle back in a couple weeks unless someone else wants to test this all out.

bhushanmeetanshi:
@TerrorSquad how to setup multi website in magento 2 using docker instance.

TerrorSquad:
> @TerrorSquad how to setup multi website in magento 2 using docker instance.

The idea from my PR was working quite well for me a few years ago when I was working with Magento 2. I see there are some conflicts now, so the implementation will need some work to make the code mergable.
If you want, fork it and try to get it work, it should be fairly straightforward to do so.

markshust:
Thanks for the work on this, @TerrorSquad, and for the honest update back in August. Going to close this one out since the repo's nginx/template structure has shifted significantly since 2022 — `compose/config/nginx/` no longer exists (templating moved to `compose/template/`), so the patch can't be rebased cleanly.

The underlying idea (multistore helper scripts + per-store nginx vhosts) is still valuable. If anyone wants to pick this up, this PR is a solid design reference — the approach of using `n98-magerun2 sys:store:list --format=csv` to drive the helpers is the right direction. A fresh PR built against the current `compose/template/` layout would be very welcome.

Closing as stale — not a reflection on the work, just the reality of a 3-year-old branch against a moved-on codebase. 🙏

PR #699, opened 2022-05-07
Fixed broken bin/cache-clean utility

bgushurst:
Updated the PHP-FPM images to install the required nodejs/npm dependencies and updated the cache-clean utility program to pass the correct parameters to start the utility. Previously incorrect params were being passed which caused the utility to fail to run.

### DIFF
--- a/compose/bin/cache-clean
+++ b/compose/bin/cache-clean
@@ -11,13 +11,13 @@ fi
 
 if [ "$1" == "--watch" ]; then
   # Kill duplicate watch process
-  WATCH_PID=$(bin/clinotty ps -eaf | grep "$CACHE_CLEAN --quiet --watch" | grep -v grep | awk '{print $2}')
+  WATCH_PID=$(bin/clinotty ps -eaf | grep "$CACHE_CLEAN -w -d /var/www/html" | grep -v grep | awk '{print $2}')
   if [[ "" !=  "$WATCH_PID" ]]; then
     bin/cliq kill -9 "$WATCH_PID"
   fi
 
   # Run watch mode in the background
-  bin/cliq $CACHE_CLEAN --quiet --watch &
+  bin/cliq $CACHE_CLEAN -w -d /var/www/html
 else
-  bin/cli $CACHE_CLEAN "$@"
+  bin/cli $CACHE_CLEAN -d /var/www/html "$@"
 fi
--- a/compose/bin/cache-clean
+++ b/compose/bin/cache-clean
@@ -1,23 +1,14 @@
 #!/bin/bash
-COMPOSER_GLOBAL=/var/www/.composer-global
-CACHE_CLEAN=${COMPOSER_GLOBAL}/vendor/bin/cache-clean.js
+CACHE_CLEAN=vendor/bin/cache-clean.js
 
 if ! bin/cliq ls $CACHE_CLEAN; then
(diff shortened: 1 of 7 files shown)


compose/bin/cache-clean
```diff
@@ -11,13 +11,13 @@ fi
 
 if [ "$1" == "--watch" ]; then
   # Kill duplicate watch process
-  WATCH_PID=$(bin/clinotty ps -eaf | grep "$CACHE_CLEAN --quiet --watch" | grep -v grep | awk '{print $2}')
+  WATCH_PID=$(bin/clinotty ps -eaf | grep "$CACHE_CLEAN -w -d /var/www/html" | grep -v grep | awk '{print $2}')
```
markshust, line 14: I'd like to retain full flag params so it's more apparent as to what is going on (`--watch` rather than `-w`, etc.).

compose/bin/cache-clean
```diff
@@ -11,13 +11,13 @@ fi
...
   if [[ "" !=  "$WATCH_PID" ]]; then
     bin/cliq kill -9 "$WATCH_PID"
   fi
 
   # Run watch mode in the background
-  bin/cliq $CACHE_CLEAN --quiet --watch &
+  bin/cliq $CACHE_CLEAN -w -d /var/www/html
 else
-  bin/cli $CACHE_CLEAN "$@"
+  bin/cli $CACHE_CLEAN -d /var/www/html "$@"
```
markshust, line 22: Why is the `-d` flag needed -- is cache-clean broken or not working without it?

compose/bin/cache-clean
bgushurst: This code has been refactored into a docker-compose run call so it's no longer applicable

compose/bin/cache-clean
bgushurst: Cache-clean is expecting to be within the magento composer vendor/bin directory and was failing to find the magento install as a result of being in the .composer-global directory without this parameter. I have refactored it to install under the main magento repo as a dev dependency to simplify this.

images/php/8.1/Dockerfile
```diff
@@ -30,6 +30,8 @@ RUN apt-get update && apt-get install -y \
     lsof \
     mailutils \
     msmtp \
+    nodejs \
+    npm \
```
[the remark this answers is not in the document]
bgushurst: This has been refactored to run on an unmodified node-slim image.

markshust:
I really appreciate the PR! Unfortunately, it seems that the cache-clean script fails because the PHP binary is not found within the Node container:

```
~/Sites/magento
❯ bin/cache-clean 
Release 1.0.47 #StandWithUkraine
20:43:26 Flushing all caches
/bin/sh: 1: php: not found
[ERROR] Command failed: php -r "echo json_encode((require '/var/www/html/app/etc/env.php') ?? []);"
/bin/sh: 1: php: not found
```

So... it appears we need PHP + Node.js in one image 🫤

If this can be figured out we can move forward with this at some point, but I don't think this will be possible at the moment. I'm going to move forward with the other commit at https://github.com/markshust/docker-magento/commit/600dd50a8ecd800370c38d03de5efcac7e2ccad9 for now, but if this ever changes, feel free to make another PR and I'll review again.

PR #707, opened 2022-05-15
bin/update: rsync ignore timestamp

msladek:
When calling `bin/update` the `rsync` call should only update actually changed files. Currently all files are overwritten because the files in `tmpupdate` will have newer timestamps. Thus disabling timestamp comparisons and adding checksumming. This shouldn't impact rsync performance since there are no large binaries in the repo. 

For more info see: https://stackoverflow.com/a/44353773/1238689


markshust:
Really neat update, didn't know about this! Thanks 👍

PR #710, opened 2022-05-19
Magento 2.4.4 as default

rafaelstz:
Since we have new patch versions and the new stable version 2.4.4 working fine. I think it's relevant to update it.
What do you think? I can change it to p3 in case you don't want the lasted version as default.


markshust:
Ah yes, this is an oversight! Sorry about that... merging in, thanks!

PR #718, opened 2022-06-16
Add install-php-extensions

rangerz:
Add `bin/install-php-extensions` for install PHP extension in php-fpm docker container

For common PHP extensions should build-in Dockerfile (image), and sometime developers may need extra PHP extension for their sites. ex: sourceguardian, ioncube_loader, ...

Ref: https://github.com/mlocati/docker-php-extension-installer

Or `install-php-extension` can build-in Dockerfile and simplify Dockerfile logic


markshust:
Appreciate the PR @rangerz!

I normally wouldn't merge this in, as it's another dependency, but I actually had trouble installing sourceguardian using the built-in method, and this worked just perfect 👌

I'm wondering if we should consider replacing the `docker-php-ext-install` commands in the main Dockerfile with this implementation instead? Do you know if there are any benefits of doing this, and/or reasons to do or not do this?

rangerz:
To @markshust 

`docker-php-extension-installer` is a script wraps `docker-php-ext-install`, properly configuring the system.
In my option, it should use in `Dockerfile`, but it has some risks we need to know before using it.
Maybe we just use in some PHP extensions only at the beginning. 

**Pros:** Simple install PHP extensions, (Enhance docker-php-ext-install)
Example 1: blackfire
```
docker-php-extension-installer blackfire # I haven't tested it but I think it could work
```
to replace some implementation details
https://github.com/markshust/docker-magento/blob/a51ae12b6f754a55b0342f3d6b2dc7e7ca8721a7/images/php/8.1/Dockerfile#L78-L84

Example 2: imagick, redis, ssh2 to help to install related libraries

**Cons:** Unofficial tool, (But it is open source)

PR #727, opened 2022-07-05
Fix user string interpreted as regex

h3xx:
This was causing issues if the volume mapped contained regex metacharacters, e.g. periods:

```sh
file='/src/p...hes'
file2='/src/patches'
IGNORE_LIST='/src/patches /src/foo'
[[ " $IGNORE_LIST " =~ $file ]] # returns true (shouldn't)
[[ " $IGNORE_LIST " =~ $file2 ]] # returns true
```
Included fix, using globs instead:
```sh
[[ " $IGNORE_LIST " = *" $file "* ]] # returns false
[[ " $IGNORE_LIST " = *" $file2 "* ]] # returns true
```

`= <glob>` is also safer than `=~ <regex>` with arbitrary user strings, but in this case, it's actually more correct.

### DIFF
--- a/compose/bin/start
+++ b/compose/bin/start
@@ -43,7 +43,7 @@ IS_VALID=true
 
 # Loop through all files missing from the docker-compose.dev.yml file
 for file in $VOLUME_LIST; do
-  if [ ! -e "$file" ] && [[ ! " $IGNORE_LIST " =~ $file ]]; then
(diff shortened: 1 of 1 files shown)


rangerz:
Approved by myself 😊😊😊
Ref: https://stackoverflow.com/questions/229551/how-to-check-if-a-string-contains-a-substring-in-bash

markshust:
Thank you, all merged!

PR #730, opened 2022-07-06
Some minor fixes for file permissions and composer auth credentials...

pniel-cohen:
I ran into some issues with file permissions and composer authentication.
Those changes made it work smoother for my case so you may consider merging if you think it can benefit others.
*I also changed the `mv .vscode src/` to `cp -r .vscode src/` just to keep the original file.


markshust:
Appreciate it, this was merged with the other merge.

PR #733, opened 2022-07-13
Added cli commands for PHP_CodeSniffer (phpcs & phpcbf)

pniel-cohen:
* Added cli commands for PHP_CodeSniffer


markshust (approved):

markshust:
@pniel-cohen appreciate it! I wasn't sure about the status of `docker compose`, but looks like it's fairly prevalent. I think Linux users may need to upgrade to the beta version of Docker Desktop, but other than that it looks pretty straight-forward, so going to merge this in. 

Thanks for the extra possible fixes around the Composer auth.json file, had a tricky time trying to fix this.... this may do it.

Issue #731, opened 2022-07-06
chmod: cannot access 'bin/magento': No such file or directory


jaywtc:
When installing Magent0 2.4.3 p1  

setup like bin/setup magento.test

it showing this error

LiamKarlMitchell:
Work-around on WSL 2 Ubuntu for me.

1. `docker exec -it magento-phpfpm-1 bash`
2. `composer create-project --repository-url=https://repo.magento.com/ magento/project-community-edition=2.4.3 .`
3. `mv project-community-edition/* .`
    Note: Not sure if needed, I may have missed the . off the end of the composer install.
4. `exit`
5. `bin/magento`

markshust:
Thanks for posting this info. I'll try to finally do an install of docker-magento on Linux after I tag & publish the next version to make sure the README is still correct. Definitely willing to take in any updates to it if needed.

LiamKarlMitchell:
Thanks Mark, if you could drop a message here when the next version is up, perhaps I can try it on weekend :)

myselfhimself:
Reproducing bug for 2.4.3-p2 with `curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/onelinesetup | bash -s -- magento.test 2.4.3-p2`. Could not reproduce fix...

Asitis:
@markshust I moved to the 2.4.4 install myself because that works fine out of the box - but I believe there is something off when loading Magento into the docker file, possibly related to permission errors that are not reproducable on macOS, [as read here](https://www.fullstaq.com/knowledge-hub/blogs/docker-and-the-host-filesystem-owner-matching-problem).
With my 2.4.4 install I also had some permission issues, that I got fixed after reading that article.

For me this happened on Ubuntu 22.04 LTS, fresh install.

myselfhimself:
> @myselfhimself can you please post your full operating system & host details? Everything works fine for me on macOS, so it'll be very hard/impossible to confirm this is even an issue without all of the debugging details.

```
Micro-Star International Co., Ltd. MS-7D25
32,0 GiB
12th Gen Intel® Core™ i7-12700K × 20
Ubuntu 22.04 LTS
64-bit
--
Docker version 20.10.17, build 100c701
```

markshust:
Are there any errors occurring when running `bin/setup`? If so and you can post the full output, we can try to find out what is going on. Manually cloning out the project within the Docker container shouldn't be needed.

myselfhimself:
[...]

PR #742, opened 2022-07-25
#731 download: adding composer allow-plugins for magento<2.4.4

myselfhimself:
…ento <2.4.4

PR #754, opened 2022-08-10
Fix for the chmod: cannot access 'bin/magento': No such file or directory issue #731

anilacikgoz:


hostep:
There's already a PR trying to fix it: https://github.com/markshust/docker-magento/pull/742

It's better because it does version checks of known Magento versions which have this problem. Which is better for the future because with Magento 2.4.4 and higher, those composer plugins are already whitelisted by default and can potentially change in the future, but that will be handled by the Magento team and shouldn't be overwritten anymore for future Magento versions by this package here.

h3xx (changes requested):
`composer update` has side effects that might break the project. In my experience,  one should never run `composer update`, only `composer install`.

markshust:
I'm going to close this PR as 2.4.3 is now EOL, and I have no plans to continue supporting any version <=2.4.3.

PR #799, opened 2022-10-25
Rename docker-compose.yml to compose.yaml (preferred)

rangerz:
Compose implementations MUST prefer canonical `compose.yaml` one.

Ref: https://docs.docker.com/compose/compose-file/#compose-file


markshust (approved):

markshust:
Appreciate this @rangerz -- didn't even know `compose.yaml` was a thing!

Since I'd probably consider this a breaking change (even though I don't think it breaks anything... can you confirm?), I'm going to hold merging this until right before the next big tagged release. That should be coming really soon.

But I reviewed everything and looks good 👍

rangerz:
I think it doesn't break anything...
Even older docker-compose should be fine because we setup the compose files in `bin/docker-compose`

I also tested my Mac (not ARM) and Linux

Mac env
Docker version 20.10.20
Docker Compose version v2.12.1

Linux env
Docker version 20.10.21
Docker Compose version v2.12.2

markshust:
Awesome, appreciate it! All merged in.

Inevix:
this tool has been used in many projects, and many internal docs have been written over the origin doc to make it more understandable for junior developers. so, all of them have to be updated "just for fun" when this renaming doesn't have any profits. thanks for taking the time

Inevix:
My point was that even small changes could break something for existing projects where this tool used. Maybe any announcements about renaming or removing of existing functionality could be a way to have time to prevent any critical consequences

PR #801, opened 2022-10-29
Improve bin/docker-compose

rangerz:
Use `docker compose` first, not `docker-compose`

Use `bin/start --no-dev` to replace `bin/docker-compose --no-dev up -d`

### DIFF
--- a/compose/bin/start
+++ b/compose/bin/start
@@ -5,6 +5,11 @@ MEM=$(docker info | grep "Total Memory" | cut -d':' -f2 | xargs | sed s/GiB//)
 # Docker reports RAM 0.2 less than what it is actually set to
 (( $(echo "$MEM < 5.7" | bc -l) )) && echo "There must be at least 6GB of RAM allocated to Docker to continue." && exit
 
+if [ "$1" == "--no-dev" ]; then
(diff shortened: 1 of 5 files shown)


compose/bin/start
```diff
@@ -36,29 +36,33 @@ function parseYaml {
   }'
 }
 
-# Check if volume files exist to avoid creating an empty folder
-VOLUME_LIST=$(parseYaml docker-compose.dev.yml services_app_volumes)
-IGNORE_LIST="./src/app/code ./src/m2-hotfixes ./src/patches ./src/var/log ./src/var/report ./src"
-IS_VALID=true
+if [ "$1" == "--no-dev" ]; then
+  bin/docker-compose --no-dev up -d --remove-orphans "$@"
+else
```
markshust, line 41: Love this update! 🧡

But rather than nesting the entire remaining code within an `else` block, can we do an early exit above after the `docker-compose` line is executed in the `--no-dev` clause?

markshust (changes requested):

compose/bin/start
rangerz: updated

markshust (approved):

PR #802, opened 2022-10-31
Use install-php-extensions to replace docker-php-ext-install

rangerz:
Ref: https://github.com/mlocati/docker-php-extension-installer

code changes: +18 −82 (Nice)


markshust:
What's the reason for doing this... just for less code? I find it a bit more confusing to find out which script supplies what package. Blackfire ini files are also gone?

rangerz:
Not only for less code, but also provide more convenient to install extra PHP extensions.

The video may update for an easier way
[Install a new PHP extension by building a custom Docker image](https://courses.m.academy/courses/set-up-magento-2-development-environment-docker/lectures/9064350)

No need `compose/bin/install-php-extensions` anymore.

I am preparing for the next pull request for this.

`env/phpfpm.env`
```
# Support list: https://github.com/mlocati/docker-php-extension-installer
PHP_EXTENSIONS=sourceguardian xxx ...
```

`image/php/8.1/docker-entrypoint.sh`
```
#!/bin/bash
set -eo pipefail

if [[ -n "$PHP_EXTENSIONS" ]]; then
  echo "Installing PHP extensions: $PHP_EXTENSIONS"
  install-php-extensions "$PHP_EXTENSIONS"
fi

exec "$@"
```

`bkackfire.ini` is generate by here
https://github.com/mlocati/docker-php-extension-installer/blob/343dc4b9dfe52571749b904d2321c0fad0c8f3b7/install-php-extensions#L2482

rangerz:
Updated

markshust:
I actually started a review of this PR, but am really hesitant to merge this in. While it does make it a lot easier to install a PHP module, it also adds a lot of complexity and "I don't know what is happening here?" logic to this project. I'm not sure it's worth the tradeoff, especially since it's not too common to install a custom PHP extensions into a Docker image (for 90% of users).

rangerz:
I agree that most users don't need to install extra PHP extensions.
Maybe `bin/install-php-extensions` is enough to add to the last line of `bin/start`

This RP simplifies the `Dockerfile` to install PHP extensions without handling their library dependencies and settings, ex: `libxxx` and `bkackfire.ini`. 
And expands `env/phpfpm.env` to install custom PHP extensions.

Here I also found that `libonig-dev` library does not need to be installed, and it's not necessary for `mbstring`

For me, [`install-php-extensions`](https://github.com/mlocati/docker-php-extension-installer) is super `docker-php-ext-install`. That's what I think it'll do (what should happen when I use it)
And support Alpine/Debian version docker images. It will be easy to switch Alpine image in the future.

Lastly, Yes, it's tradeoff. Maybe there's a better way that we don't know yet.

PR #830, opened 2022-12-09
Add SSL CA integration support for Firefox Developer Edition

h3xx:
This makes it easier to add more Firefox flavors to the list.

Also spruce up script:
- Don't do `[[ -d ]]`, then `mkdir -p` -- just use `mkdir -p` (Unfortunately, OSX's `install` doesn't support `-D`)
- Add more messages about what it's doing.


rangerz:
Is importing CA into Firefox still working for MacOSX ?

PR #835, opened 2022-12-22
Mage-OS Mirror Support

pykettk:
# Description
Add support for installing Magento via [all of the currently available Mage-OS mirror repositories](https://mage-os.org/distribution#magento-mirrors).

### DIFF
--- a/compose/bin/download
+++ b/compose/bin/download
@@ -12,8 +12,14 @@ bin/setup-composer-auth
 
 bin/fixowns
 
-if [ "$EDITION" == "mageos" ]; then
+if [ "$EDITION" == "mageos-nightly" ]; then
   bin/clinotty composer create-project --stability alpha --repository-url=https://upstream-nightly.mage-os.org magento/project-community-edition .
+elif [ "$EDITION" == "mageos-mirror" ]; then
+  bin/clinotty composer create-project --repository-url=https://mirror.mage-os.org magento/project-community-edition="${VERSION}" .
+elif [ "$EDITION" == "mageos-hypernode-mirror" ]; then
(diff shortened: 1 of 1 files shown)


markshust:
Thanks @pykettk, this looks nice! Can we also retain a `mageos` option that aliases back to one of these mirrors? I'm not sure which is the best to default to.

markshust:
Also see shellcheck output above 👆 This just checks to ensure formatting is consistent throughout the project.

Summary:
```
Notice: ./compose/bin/setup-composer-auth:14:6: note: Double quote to prevent globbing and word splitting. [SC2086]
Notice: ./compose/bin/setup-grunt:63:16: note: Double quote to prevent globbing and word splitting. [SC2086]
Notice: ./compose/bin/setup-grunt:64:16: note: Double quote to prevent globbing and word splitting. [SC2086]
```

pykettk:
@markshust Keeping a `mageos` option is also a good idea 🙂 I think I'd be leaning towards aliasing it to the `mageos-mirror` option as that's the most "official" mirror (if such a thing exists 😛) and is guaranteed to work unlike the `mageos-nightly` option.

Happy to make this change, will be a quick one, just confirm your preference 🙂

pykettk:
RE: the spellchecker - as the reported issues are not connected to the changes I'm making here,  I have opened a new pull request to address (#840) those, @markshust 👍

markshust (approved):

markshust:
Great, I'll just go along with your suggestion! Merged in 👍

pykettk:
Thanks, @markshust :)

PR #836, opened 2022-12-22
Create "Manual" Setup Script

pykettk:
# Description
Reduces the number of steps required to perform manual setup and updates the documentation accordingly.

This is quite an opinionated change so definitely invite input and discussion around these changes 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,20 +159,17 @@ cd $_
 # Download the Docker Compose template:
 curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/template | bash
 
-# Download the version of Magento you want to use with:
-bin/download 2.4.5-p1 community
-# You can specify the version and type (community, enterprise, or mageos).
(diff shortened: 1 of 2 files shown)


markshust:
Yea I'm a bit biased against merging this in because the documentation for manual setup sort of helps everyone also understand what's going on under the hood.

I can still easily understand it, but I'm not sure this level of abstraction is needed, especially since anyone running this will only ever do this once per project.

I'll keep this open so others can chime in, but leaning towards closing this one out.

pykettk:
That's a fair point. I guess it's a balance between convenience and education - I'd probably favour the latter honestly. Happy for this to be closed now, people can still comment on it and we can have discussions etc. while keeping the repository clean 🙂

markshust:
Ok yea sure sounds good, I always hear everyone's feedback out even after issues/PRs are closed. Nothing is ever final 😄

PR #837, opened 2022-12-22
Stop Containers Before Removing

pykettk:
# Description
Remove the need for manually stopping all containers before removing them when running `bin/removeall`.

Another opinionated change so input and discussion is welcome 🙂


markshust:
Thanks! I totally agree with this, there are no downsides 😃

PR #846, opened 2023-01-09
Enable xdebug to listen to Magento CLI commands

massalinux:
Hi Mark, i added this script because often i need to debug some console commands and i used to edit the env/phpfpm.env   and roll it back manually.

Let me know if you're interested on that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ It is recommended to keep your root docker config files in one repository, and y
 - `bin/devconsole`: Alias for `bin/n98-magerun2 dev:console`
 - `bin/docker-compose`: Support V1 (`docker-compose`) and V2 (`docker compose`) docker compose command, and use custom configuration files, such as `compose.yml` and `compose.dev.yml`
 - `bin/download`: Download specific Magento version from Composer to the container, with optional arguments of the version (2.4.5-p1 [default]) and type ("community" [default], "enterprise", or "mageos"). Ex. `bin/download 2.4.5-p1 enterprise`
+- `bin/debug-cli`: Enable xdebug for Magento Console Command. Second argument is optional, default is PHPSTORM Ex. `bin/debug-cli enable [PHPSTORM]`
 - `bin/fixowns`: This will fix filesystem ownerships within the container.
 - `bin/fixperms`: This will fix filesystem permissions within the container.
 - `bin/grunt`: Run the grunt binary. Ex. `bin/grunt exec`
--- a/compose/bin/debug-cli
+++ b/compose/bin/debug-cli
@@ -1,0 +1,61 @@
+#!/bin/bash
+S=$(bin/clinotty cat /usr/local/etc/php/php.ini | grep -iGc 'xdebug.mode = off');
+
+if [[ $S == 1 ]]; then
+  echo "Enable xdebug first"
+  exit 1
+fi
+
+R=$(cat ./env/phpfpm.env | grep -c 'XDEBUG_CONFIG=idekey')
+
+if [[ -z "$2" ]]
+  then
+    platform="PHPSTORM"
+  else
+    platform=$2
+fi
+
+local_debug_status(){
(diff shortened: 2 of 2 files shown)


compose/bin/debug-cli
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+S=$(bin/clinotty cat /usr/local/etc/php/php.ini | grep -iGc 'xdebug.mode = off');
+
+if [[ $S == 1 ]]; then
+  echo "Enable xdebug first"
```
h3xx, line 5: Why not run `bin/xdebug enable` if you've determined it needs to be done? You can remove the check entirely since `bin/xdebug` itself will detect it's enabled?

massalinux: @h3xx yes of course, maybe could be also as an additional argument of bin/xdebug like --cli

markshust: @massalinux I think he means instead of running the line:

    echo "Enable xdebug first"

Just run

    bin/xdebug enable

This way rather than prompting the user to enable Xdebug, you're just enabling it.

README.md
```diff
@@ -251,6 +251,7 @@ It is recommended to keep your root docker config files in one repository, and y
 - `bin/devconsole`: Alias for `bin/n98-magerun2 dev:console`
 - `bin/docker-compose`: Support V1 (`docker-compose`) and V2 (`docker compose`) docker compose command, and use custom configuration files, such as `compose.yml` and `compose.dev.yml`
 - `bin/download`: Download specific Magento version from Composer to the container, with optional arguments of the version (2.4.5-p1 [default]) and type ("community" [default], "enterprise", or "mageos"). Ex. `bin/download 2.4.5-p1 enterprise`
+- `bin/debug-cli`: Enable xdebug for Magento Console Command. Second argument is optional, default is PHPSTORM Ex. `bin/debug-cli enable [PHPSTORM]`
```
markshust, line 254: This line should to follow the format of the rest of the documentation, something like:

- `bin/debug-cli`: Enable Xdebug for bin/magento, with an optional argument of the IDE key. Defaults to PHPSTORM Ex. `bin/debug-cli enable PHPSTORM`

compose/bin/debug-cli
```diff
@@ -0,0 +1,61 @@
...
+  local_debug_disable
+elif [[ $firstArgLetter == "e" ]]; then
+  local_debug_enable
+elif [[ $firstArgLetter == "t" ]]; then
+  local_debug_toggle
+elif [[ $firstArgLetter == "s" ]]; then
+  local_debug_status
+else
+    printf "Please specify either 'disable', 'enable', 'status' or 'toggle' as mandatory argument.\nSpecify as an optional second argument the platform. Default is PHPSTORM\nEx: bin/debug-cli enable [PHPSTORM]\n"
+fi
```
markshust, line 61: Needs a newline at the end of this file

markshust:
Thanks! Left a few comments. Note that shellcheck is also failing with:

```
Notice: ./compose/bin/debug-cli:9:9: note: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead. [SC2002]
```

compose/bin/debug-cli
h3xx: I implied removing lines 2-5 entirely including the check (`S=$(...)`) and replace with something like:
```sh
# Ensure xdebug is enabled
bin/xdebug enable
```

massalinux: yes of course, what i meant was that you can actually incorporate this functionality directly inside the bin/xdebug command as on optional parameter instead of create a new command.

Meanwhile i've modified as you suggested the command. I'll add the suggestion from @markshust and update the PR.

markshust (approved):

markshust:
Thanks so much for the updates... all merged in! 🎉

massalinux:
🚀

PR #848, opened 2023-01-12
Fix failing "npm install" of puppeteer due to missing Chromium

h3xx:
This fix sets env inside the `phpfpm` container `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true`.

Solution taken from this comment by @talisto: https://github.com/puppeteer/puppeteer/issues/7740#issuecomment-970490323

Note for reproduction:
- You must copy `package.json.sample` to `package.json` and you should remove `package.lock` before trying to reproduce. (`cd src && cp package.json.sample package.json && rm package.lock`)
- You should also make sure you're running in an AMD64 container; `bin/cli uname -m` should return `aarch64`
- If `bin/npm i` fails, the bug persists

When running on an AMD64 container (`phpfpm`), `npm install` would fail when installing puppeteer with the following error:

```
npm ERR! code 1
npm ERR! path /var/www/html/node_modules/puppeteer
npm ERR! command failed
npm ERR! command sh -c node install.js
npm ERR! The chromium binary is not available for arm64.
npm ERR! If you are on Ubuntu, you can install with:
npm ERR!
npm ERR!  sudo apt install chromium
npm ERR!
npm ERR!
npm ERR!  sudo apt install chromium-browser
npm ERR!
npm ERR! /var/www/html/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:115
npm ERR!                     throw new Error();
npm ERR!                     ^
npm ERR!
npm ERR! Error
npm ERR!     at /var/www/html/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:115:27
npm ERR!     at FSReqCallback.oncomplete (node:fs:204:21)
npm ERR!
npm ERR! Node.js v18.2.0
```


markshust:
I appreciate it! This is just an env-var update and shouldn't have any downsides, so I'm just going to merge it in. If something isn't right afterwards, just comment back or open up a new issue. Thanks!

PR #855, opened 2023-01-26
Simplify RAM check

h3xx:
Use `docker info -f` to get the number of bytes. This results in fewer dependencies on external programs, and better usage of Bash shell math.

Removes comment stemming from confusion about GiB vs GB. :-)

### DIFF
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -1,9 +1,12 @@
 #!/bin/bash
 set -o errexit
 
-MEM=$(docker info | grep "Total Memory" | cut -d':' -f2 | xargs | sed s/GiB//)
-# Docker reports RAM 0.2 less than what it is actually set to
-(( $(echo "$MEM < 5.7" | bc -l) )) && echo "There must be at least 6GB of RAM allocated to Docker to continue." && exit
+MEM_BYTES=$(docker info -f '{{.MemTotal}}')
+MEM_MB=$(( MEM_BYTES / 1000000 ))
+(( MEM_MB >= 6000 )) || {
+  echo 'There must be at least 6GB of RAM allocated to Docker to continue.' >&2
+  exit 1
+}
 
 # shellcheck source=../env/db.env
 source env/db.env
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -3,10 +3,8 @@ set -o errexit
 
 MEM_BYTES=$(docker info -f '{{.MemTotal}}')
 MEM_MB=$(( MEM_BYTES / 1000000 ))
-(( MEM_MB >= 6000 )) || {
-  echo 'There must be at least 6GB of RAM allocated to Docker to continue.' >&2
-  exit 1
-}
+# When Docker Desktop is set to 6GB in the GUI, it is reported as 6227 MB
+(( MEM_MB < 6227 )) && echo "There must be at least 6GB of RAM allocated to Docker in order to continue." && exit
 
 # shellcheck source=../env/db.env
 source env/db.env


markshust:
I like the addition of these two lines:

```
MEM_BYTES=$(docker info -f '{{.MemTotal}}')
MEM_MB=$(( MEM_BYTES / 1000000 ))
```

But I think the if check is now quite a more confusing. Can we bring back the old one liner, but using the two vars above?

Also, notice the previous comment about Docker misreporting mem. It should fail if the mem is less than 5.7GB, not 6GB. I'm not sure why Docker does this, but setting it to 6GB does cause the memory to be incorrectly identified for some reason.

Thanks!

compose/bin/setup
```diff
@@ -1,9 +1,12 @@
 #!/bin/bash
 set -o errexit
 
-MEM=$(docker info | grep "Total Memory" | cut -d':' -f2 | xargs | sed s/GiB//)
-# Docker reports RAM 0.2 less than what it is actually set to
-(( $(echo "$MEM < 5.7" | bc -l) )) && echo "There must be at least 6GB of RAM allocated to Docker to continue." && exit
+MEM_BYTES=$(docker info -f '{{.MemTotal}}')
+MEM_MB=$(( MEM_BYTES / 1000000 ))
+(( MEM_MB >= 6000 )) || {
```
markshust, line 6: /wtd use previous one-liner logic instead

h3xx: Thanks bot, but that change introduces a syntax error.

h3xx:
> Also, notice the previous comment about Docker misreporting mem. It should fail if the mem is less than 5.7GB, not 6GB. I'm not sure why Docker does this, but setting it to 6GB does cause the memory to be incorrectly identified for some reason.

I think you're confusing [GB and GiB](https://duckduckgo.com/?t=ffab&q=gib+vs+gb&ia=answer). Briefly, GiB is base-1000, GB is base 1024. GiB * 1,000,000,000 = number of bytes, GB * 1024 * 1024 * 1024 = number of bytes.

The Docker Desktop interface is using a different unit from the `docker info` (without `-f`):
- `docker info` without `-f` gives it back in GiB, which again is base 1000.

`docker info -f '{{.MemTotal}}'` gives it back in the number of bytes, and since I divide it by exactly 1 million, the `MEM_MB` is in **MiB**. (TODO change variable to `MEM_MIB`)

When Docker Desktop is set to use 6GB exactly:
- Original:
  - `MEM` is set to `5.8`
  - `MEM` is more than or equal to `5.7` --> Check passes (should ✅)
- My change:
  - `docker info -f '{{.MemTotal}}'` reports `6227464192`
  - `MEM_MB` under my change gets set to `6227`.
  - `MEM_MB` more than or equal to 6000 --> Check passes (should ✅)

When Docker Desktop is set to use 5.8GB:
- Original:
  - `MEM` is set to `5.554`
  - `MEM` is less than `5.7` --> Check does not pass (shouldn't ✅)
- My change:
  - `docker info -f '{{.MemTotal}}'` reports `5963747328`
  - `MEM_MB=5963`
  - `MEM_MB` is less than 6000 --> Check does not pass (shouldn't ✅)

h3xx:
There *is* something fishy going on with Docker. 6,227,464,192 is *not* exactly 6GB. It's 214,986,752 bytes short of 6GB.

compose/bin/setup
```diff
@@ -1,9 +1,12 @@
...
+  echo 'There must be at least 6GB of RAM allocated to Docker to continue.' >&2
+  exit 1
+}
```
markshust, line 9: /wtd use previous one-liner logic that checks the memory and outputs the message rather than the multi-line version

markshust:
Yup hence the 5.6GB check 😁 I removed the wtd bot updates, that tool doesn't seem to be working well for handling rewriting of diffs.

markshust:
I made some updates to use this new logic and the previous one-liner, as it's a bit simpler & easier to understand.

PR #863, opened 2023-02-18
Correct base url in echo

danstaver:
Small fix for 'web/secure/base_url' being mentioned twice, when it is ssecure and insecure being set


markshust:
Thanks for this typo fix!

PR #869, opened 2023-03-01
Add support for multi-storefront Magento

h3xx:
**I really want to know if the method described under `How multi-storefront switching works` is the standard way of switching between storefronts, or if it's just a technique peculiar to my workplace.**

This *may* be going too far towards how one hosting solution (Adobe's Magento Cloud service) sets things up; other hosting solutions are available.

## My change

(Aside from a slight refactor to make it easier) this change adds an additional YAML file, `compose.multi-storefront.yaml`, that adds 2 additional volumes to `app` and `phpfpm`:
- `src` -> `/app` - this mimics the directory layout when deployed to Magento Cloud.
- `src/php.ini` -> `/usr/local/etc/php/conf.d/local-php.ini` - mapped so that PHP will load it.

It *only* uses `compose.multi-storefront.yaml` if both `src/php.ini` and `src/magento-vars.php` exist.

## How multi-storefront switching works

Multi-storefront switching is performed by modifying the `$_SERVER['MAGE_RUN_CODE']` and `$_SERVER['MAGE_RUN_TYPE']` values inside of `$_SERVER`. It's done in a special file called `magento-vars.php`. Here's an example:

```php
<?php
/**
 * Copyright © Magento, Inc. All rights reserved.
 * See COPYING.txt for license details.
 */

/**
 * Enable, adjust and copy this code for each store you run
 *
 * Store #0, default one
 *
 * if (isHttpHost("example.com")) {
 *    $_SERVER["MAGE_RUN_CODE"] = "default";
 *    $_SERVER["MAGE_RUN_TYPE"] = "store";
 * }
 *
 * @param string $host
 * @return bool
 */
function isHttpHost(string $host)
{
    if (!isset($_SERVER['HTTP_HOST'])) {
        return false;
    }
    return $_SERVER['HTTP_HOST'] === $host;
}
/**
 * Enable, adjust and copy this code for each store you run
 *
 * Store #0, default one
 */
if (
    isHttpHost("www.defaultstore.example.com")
    || isHttpHost("defaultstore.example.com")
    || isHttpHost("defaultstore.example.dev")
) {
    $_SERVER["MAGE_RUN_CODE"] = "default";
    $_SERVER["MAGE_RUN_TYPE"] = "store";
}
if (
    isHttpHost("www.store1.example.com")
    || isHttpHost("store1.example.com")
    || isHttpHost("store1.example.dev")
) {
    $_SERVER["MAGE_RUN_CODE"] = "store1_store_view";
    $_SERVER["MAGE_RUN_TYPE"] = "store";
}
```

This file is included in every request via `src/php.ini`:
```ini
;
; Multi store support
;
auto_prepend_file = /app/magento-vars.php
```

Using our same example domains, getting this setup for local development depends on modifications to `/etc/hosts` to make both `defaultstore.example.dev` and `store1.example.dev` work in the browser:
```hosts
127.0.0.1 ::1 defaultstore.example.dev
127.0.0.1 ::1 store1.example.dev
```

Running `bin/setup-ssl defaultstore.example.dev store1.example.dev` is not a bad idea either.

### DIFF
--- a/compose/bin/docker-compose
+++ b/compose/bin/docker-compose
@@ -6,8 +6,24 @@ else
   DOCKER_COMPOSE="docker-compose"
 fi
 
-if [ "$1" == "--no-dev" ]; then
-  ${DOCKER_COMPOSE} -f compose.yaml -f compose.healthcheck.yaml "${@:2}"
-else
-  ${DOCKER_COMPOSE} -f compose.yaml -f compose.healthcheck.yaml -f compose.dev.yaml "$@"
+DOCKER_COMPOSE_YAMLS=(
+  compose.yaml
+  compose.healthcheck.yaml
+)
+
+if [[ -f src/php.ini && -f src/magento-vars.php ]]; then
+  DOCKER_COMPOSE_YAMLS+=(compose.multi-storefront.yaml)
 fi
+if [[ $1 = '--no-dev' ]]; then
+  shift 1
+else
+  DOCKER_COMPOSE_YAMLS+=(compose.dev.yaml)
+fi
+
+DOCKER_COMPOSE_ARGS=()
+for YAML in "${DOCKER_COMPOSE_YAMLS[@]}"; do
+  DOCKER_COMPOSE_ARGS+=(-f "$YAML")
+done
+${DOCKER_COMPOSE} \
+  "${DOCKER_COMPOSE_ARGS[@]}" \
+  "$@"
--- a/compose/compose.multi-storefront.yaml
+++ b/compose/compose.multi-storefront.yaml
@@ -1,0 +1,12 @@
+version: "3"
+
+services:
+  app:
+    volumes: &appvolumes
+      # Magento Cloud-like path support
+      - ./src:/app:cached
+      # Multi-storefront support depends on "auto_prepend_file = /app/magento-vars.php" being in src/php.ini
+      - ./src/php.ini:/usr/local/etc/php/conf.d/local-php.ini:cached
+
+  phpfpm:
+    volumes: *appvolumes


piotrkwiecinski:
This is not needed to run multistore. I managed to have running instance with adding a file for mapping a domain to run code:

I have a file `store.map.conf` with content:
```
map $http_host $MAGE_RUN_CODE {
    sub1.domain.loc sub1_store_code;
    sub2.domain.loc sub2_store_code;
    domain.loc default;
}
```

In `docker-compose.dev.yaml`:
```
services:
  app:
    volumes: &appvolumes
      - ./store.map.conf:/etc/nginx/conf.d/store.map.conf:cached
```
@markshust what do you think about adding readme for this instead of the PR.

The proposed solution isn't generic as `magento-vars.php` in cloud projects applied environment specific domains based on a branch you're on. It's used for mapping domains for environments on platform.sh not your local environment.

markshust:
@piotrkwiecinski I actually kinda like this PR. So often, I hear from devs that they do not know how to make updates to the Docker image/config and there are a few steps in play.

@h3xx so sorry for the delay, I've not been attentive to OSS lately. I like this PR.

To clarify, one would need to just create `src/magento-vars.php` and `src/php.ini` files to get the multi-store/domain support?

Also, does just adding `src/php.ini` a file work as expected, applying this update to the PHP config? This would allow someone to easily modify the PHP config without messing with the containers. I think just having a `src/php.ini` file to apply PHP config would make it so easy to modify this.

markshust:
Thanks again @h3xx for the original proposal here, and @piotrkwiecinski for the nginx `map` alternative — both approaches are genuinely useful, and the discussion on this PR is what made that clear.

Rather than baking either pattern into `compose/bin/docker-compose`, I'd like to move forward with documenting both. A couple of things have changed since this was opened:

- #1397 added support for a `compose.override.yaml` file that `bin/docker-compose` automatically includes if present, and which `bin/update` won't overwrite. That means anything this PR's `compose.multi-storefront.yaml` would have done can now be expressed as a 6-line override file the user drops in themselves — no script branching, no auto-detection of `src/php.ini` + `src/magento-vars.php`, no extra compose file shipped in the repo.
- Keeping multi-storefront setup out of the shipped scripts also avoids the "this only matches Adobe Commerce Cloud's layout" concern @piotrkwiecinski raised — other hosting setups (and the nginx `map` approach) get equal billing.

I've opened #1430 with a docs section covering both patterns, which I'll merge in lieu of this. Going to close this one out, but really appreciate the patience on the long review cycle — the conversation here directly shaped what landed.

markshust:
Closing in favor of #1430 (docs-only). Thanks again to everyone who contributed to the discussion here.

PR #874, opened 2023-03-06
Setup script is divided to 2 parts

mucan54:
Sometimes we just need to run bin/magento setup:install script, so maybe it can be a better approach to divide the setup script to can use the setup-install command alone.


markshust:
I like this update as it just makes everything more simple! All good, appreciate it!! 👍

Issue #864, opened 2023-02-21
bin/magento doesn't find /bin/bash


daef:
On systems which don't have bash under /bin/bash the scripts are broken.

This could probably be fixed by the superior shebang `#!/usr/bin/env bash` ([more](https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html)).

daef:
I use [nixos](https://nixos.org/), which doesn't keep binaries around in /bin because one might have different versions installed.
/usr/bin/env is POSIX

markshust:
I really appreciate it @daef! All merged in 🤓

PR #879, opened 2023-03-14
use /usr/bin/env to discover bash location

daef:
As discussed in Issue https://github.com/markshust/docker-magento/issues/864

### DIFF
--- a/compose/Makefile
+++ b/compose/Makefile
@@ -1,4 +1,4 @@
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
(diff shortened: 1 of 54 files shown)


markshust:
I really, really appreciate this PR @daef! If there are any side-effects from merging this in, just lmk.

sagar1dahiwala:
after the installation when I try to run `make bash` got this error `bash: ./usr/bin/env: No such file or directory` on Mac. I see the bin/bash file seems to be corrupted. I have to rename the `bash` file to `bash.sh` and update the make file to fix it. 
`bash:
	@./bin/bash.sh`

PR #900, opened 2023-04-06
Modify RabbitMQ credentials in test templates to comply with rabbitmq.env

jbrada:
When running the integration tests, it fails to connect to RabbitMQ because the pre-made integration tests configuration templates contain different RabbitMQ login credentials than are in rabbitmq.env.

This pull request modifies the templates to contain the same values as the default rabbitmq.env values


piotrkwiecinski (approved):

markshust:
Thanks so much for this update @jbrada and for the approval @piotrkwiecinski! I've been down a bit of a wormhole this year but am making my way back to this project to get some updates in.

Merging this in 👍

PR #901, opened 2023-04-06
Tweak version detection in setup-integration-tests script

jbrada:
I've made some changes to the `setup-integration-test` script that addresses an issue where it was not possible to use the script when a specific configuration of Xdebug was in place. 

With this pull request, I've tweaked the version detection to evaluate only rows with the Magento string. The tweak should help to resolve the issue and make the script more reliable overall.

(this is an example output of `bin/magento --version --no-ansi` when xdebug is enabled on each request and disabled PHPStorm's Listening for Debug Connection)
image

Thank you @markshust for your hard work in maintaining this package!


piotrkwiecinski (approved):
Looks good

markshust:
This script has been updated since, and unfortunately, grepping out Magento no longer works.

We have changed to this line:

```
bin/magento-version | cut -d'.' -f1,2
```

If the output of `bin/magento-version` is `Magento CLI 2.4.8-beta2`, this line will return `2.4`.

I'm hesitant to do anything in regards to this, namely because Xdebug throwing warnings is telling you something is wrong or needs to be fixed. I don't think this edge-case is worth fixing (better off fixing the Xdebug issue, as that is very non-standard and not the expected output when running `bin/magento-version`. Going to close this one out.

PR #933, opened 2023-05-21
Update setup-install

mucan54:
Disable Two factor aut as default on setup.


markshust:
I don't really care for this update, as this disables 2FA by default, which could open up security holes.

FYI I did create a module to disable 2FA in dev. You can grab it at https://github.com/markshust/magento2-module-disabletwofactorauth

I don't install the module by default as I don't agree with force-feeding fresh installs with any third-party modules (even my own), but you can easily set this up for a new project to disable 2FA.

PR #934, opened 2023-05-27
Disable Locking for Redis of session save

rangerz:
Improving Magento 2 performance to enable `--session-save-redis-disable-locking=1`
https://objectsource.co.uk/magento-2-performance-issue-with-sessions/

Also add `--search-engine=opensearch` for installing Magento 2.4.6


compose/bin/setup-install
```diff
@@ -37,16 +37,18 @@ bin/clinotty bin/magento setup:install \
...
+  --cache-backend-redis-db="$CACHE_BACKEND_REDIS_DB" \
+  --page-cache="$PAGE_CACHE" \
+  --page-cache-redis-server="$PAGE_CACHE_REDIS_SERVER" \
+  --page-cache-redis-db="$PAGE_CACHE_REDIS_DB" \
+  --session-save="$SESSION_SAVE" \
+  --session-save-redis-host="$SESSION_SAVE_REDIS_HOST" \
+  --session-save-redis-log-level="$SESSION_SAVE_REDIS_LOG_LEVEL" \
+  --session-save-redis-db="$SESSION_SAVE_REDIS_DB" \
+  --session-save-redis-disable-locking="$SESSION_SAVE_REDIS_DISABLE_LOCKING" \
+  --search-engine=opensearch \
```
piotrkwiecinski, line 51: Adding `opensearch` shouldn't be needed as this is a default engine in 2.4.6. It would have to be explicit if you would want to switch back to `elasticsearch7`

piotrkwiecinski:
Is there visible improvement with disabled lock while working locally? Have you been able to measure it somehow?

rangerz:
> Is there visible improvement with disabled lock while working locally? Have you been able to measure it somehow?

I didn't find a visible way to measure it in the local development environment... lol

markshust:
If it ain't broke... 😅. I'm hesitant to pull this in without a reason for doing so.

PR #956, opened 2023-07-05
bin/docker-start helper script

nkajic:
Hi @markshust
this is just an idea. Please see if it works for you and your project.

bin/docker-start can be used as standalone command to start docker (currently mac only) or can be used inside other automatisation scripts. I use it in another script which recreate local projects from project source and database dumps.

Greetings, 
Neven


markshust:
This is interesting, I like it. Though I use OrbStack, we can check to see if that is installed first and if not, then fallback to Docker Desktop. I think this script needs some additions possibly for Linux as well to check if the docker daemon is started and if not, start it there as well.

markshust:
Thanks for this! I updated it to also detect OrbStack as well. Neat :)

PR #963, opened 2023-07-19
Changed sequence of commands

sfritzsche:
Moved the command "bin/mysqldump", since this is only available for a new "docker-magento" user after downloading the Docker Compose template.


markshust:
Thanks for this, good catch 👍

PR #979, opened 2023-08-17
Remove network along with removeall

dk4software:
Removeall is not removing the network. And after some time, it gives error and you will have to manually remove some of the unused networks:
⠿ Network magento-local-setup-test_default  Error                                                                                                                      0.0s
failed to create network magento-local-setup-test_default: Error response from daemon: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network
Failed to start Docker services


markshust:
Thank you, good idea. I think we need to add a line break at the end of the `compose/bin/removenetwork` file though to be consistent with the rest of the files. As soon as that is done, I'll go ahead and merge it in.

PR #981, opened 2023-08-23
Update copyfromcontainer to create src folder

rafaelstz:
It creates the src folder when copying all the code from the repo first the first time, by the src doesn't exist. 

An example is when the dev follows the README part saying to run bin/cli git clone .... to start the project.


markshust:
Thanks for this, I don't see any downsides so merging this in

PR #982, opened 2023-08-23
Update setup-install to cleanup

rafaelstz:
The usage of `--cleanup-database` helps to finish the installation in most of cases, avoiding small fixable issues during clean installations and installations with pre-created databases as well.


markshust:
I didn't know about this flag... thanks so much! Merging in 👍

PR #998, opened 2023-10-17
Add Optional Support for Cloudflare Tunnels

YPyltiai:
This MR adds support for Cloudflare Tunnels (https://www.cloudflare.com/products/tunnel/).
This is helpful when you need your instance to have public domain, so you can allowlist it with 3-p services that don't allow connections from localhost.

**Added:**
- Main config update (commented Cloudflare Tunnel section)
- Env file for Cloudflare to include your tunnel's token
- Update to Readme to cover setup part

**NOTE:** Cloudflare Tunnel makes app visible to the world, so use it carefully to not leave Magento instance always available publicly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,6 +582,19 @@ Next, open up the `bin/start` helper script and uncomment the line:
 
 Finally, restart the containers with `bin/restart`. After doing so, everything is now configured and you can use a browser extension to profile your Magento store with Blackfire.
 
+### Cloudflare Tunnel
+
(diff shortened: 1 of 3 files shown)


YPyltiai:
Hi @markshust ! Any chance you could check this out?
It is a totally optional feature but it enables many use cases to be supported out of the box.
Thanks!

piotrkwiecinski (approved):

markshust:
Thanks so much for this (and for your patience), merging it in!

YPyltiai:
Thanks for taking time to review & merge, Mark!

PR #1001, opened 2023-10-26
Update default Magento version to 2.4.6-p3.

cosmicvibes:


markshust:
Thanks so much for this PR! Looks great, merging it in 👍

PR #1015, opened 2023-12-04
Add edit-hosts script

YevhenZvieriev:
This script is created to simplify editing the /etc/hosts file for local development purposes. It grants an easy way to modify the host's file without directly using sudo or manually navigating to the file. Developers can execute this script to open the host file in the nano text editor, allowing them to add or modify entries as needed. The script checks for administrator rights and prompts for a password if necessary, ensuring proper permissions for editing the system host file.

### DIFF
--- a/compose/bin/edit-hosts
+++ b/compose/bin/edit-hosts
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Set the default editor to nano
+EDITOR=nano
+
+# Check if nano is installed; if not, use vim
+if ! command -v $EDITOR &> /dev/null; then
+  EDITOR=vim
+fi
+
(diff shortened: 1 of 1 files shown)


markshust:
Thank you for the PR. Although after thinking about this, it feels like an over-complication. I think just running `sudo vi /etc/hosts` is simpler & easier, and requires no maintenance? I think what I'm saying is that I'm failing to understand the reason for this script.

YevhenZvieriev:
Hi, Mark!

Thank you for your feedback. I understand your position. You're probably right, the user doesn't need to edit the `/etc/hosts` file so often that it's worth creating a separate script for this.

I close this PR.

PR #1016, opened 2023-12-04
Add extra-settings script

YevhenZvieriev:
This script is designed to simplify the configuration of the Docker environment and related settings. The primary tasks it addresses are:

1. Docker Container IP Address:
   - Adds the container's IP address to the /etc/hosts file for convenient access.

2. Opening Port 9003 for Xdebug:
   - Provides the option to open port 9003 for Xdebug.

3. Increasing Virtual Memory Map Count for Elasticsearch:
   - Allows increasing the virtual memory map count for Elasticsearch if needed.

### User Instructions:

1. Execute the script to perform the necessary configurations.

2. Follow the prompts and choose options according to your preferences.

3. Verify that all tasks have been completed successfully.

Please feel free to reach out if you have any questions or suggestions for improving the script.

Screenshot_2023-12-04_17-00-45


markshust:
I like the ideas in this PR. However, I don't really like how this is in a non-descript "extra-settings" file. It feels a bit out of place & random.

Does it make sense to rename this to `bin/configure-linux`? I think the host.docker.internal ip is automatically taken care of for you in macOS, and the `iptables` command doesn't exist (or is needed) on mac. This would also open up this script for possible future enhancements that are specific to Linux.

And I don't think we need the elasticsearch command at all, because I _think_ we can set this memory param as an environment variable at https://github.com/markshust/docker-magento/blob/master/compose/compose.yaml#L63. If so, it would make sense to just add this `max_map_count` prop as a commented-out setting under that `environment` property right in the `compose.yaml` file, along with a comment-out description for what it does.

YevhenZvieriev:
Hi, Mark!

Thank you for your feedback. 
Good idea! 

1. I've renamed the script name from `bin/extra-settings` to `bin/configure-linux`, and modified the script.

2. I've added the `max_map_count` parameter for Elasticsearch and OpenSearch with comments to the `compose.yaml` file.

markshust:
I made a few updates to make the script more portable, consistent with the logic within other scripts, and simplify the logic.

PR #1017, opened 2023-12-07
Add container monitoring script

YevhenZvieriev:
This Bash script provides a simple and clear display of Docker container statistics, including CPU usage and memory usage. The script utilizes the docker stats command and formats the output into a tabular format for easy readability.

Screenshot_2023-12-07_02-23-24


rangerz:
How about this way ? More concise I think.
```bash
docker stats $(bin/docker-compose ps -q)
```

EvgeniyZverev:
@rangerz Exactly!

I thought about that, but in my viewpoint, docker stats was giving too much information so I reduced it to just name, CPU usage, memory usage, and memory limit, but if this command does it in one line, why not?

I took that into account
Please feel free to view my other scripts.

markshust (approved):

markshust:
I really like the initial PR. I renamed it to `bin/docker-stats` though as that follows the naming convention of the underlying Docker command that we are running. Fixed the shellcheck error and added it to the readme.

YevhenZvieriev:
Hi, Mark!

I would like to draw your attention to the fact that the script has been updated after @rangerz's suggestion, and now it displays 

CONTAINER ID   NAME                       CPU %     MEM USAGE / LIMIT    MEM %     NET I/O       BLOCK I/O         PIDS

I think that we should modify the `README.md` file, isn`t we?

PR #1030, opened 2023-12-26
Update php.ini

emAfzal52:


markshust:
I think 100M seems enough for me. These images are meant strictly for development... I can't see someone testing out a 100M+ file in development, or why they would want to do this.

PR #1040, opened 2024-01-11
Added script for creating Magento admin user and customer

YevhenZvieriev:
Added a script ` bin/create-user ` to automate the creation of a Magento admin user and customer. The script utilizes the `bin/magento admin:user:create` command with predefined values for the admin user, and n98 for creating a customer.

admin-user-create


YevhenZvieriev:
By the way, I can modify this script so that the user can create an admin user or new customer account.

YevhenZvieriev:
Hi, @markshust

Good idea!

I`ll do it in my next commit.

YevhenZvieriev:
## Updated 25.02.2024:

**Admin User Creation**: The script includes a function to create an admin user account, allowing customization of username, email, first name, last name, and password.

**Customer Account Creation:** Another function enables the creation of customer accounts, with the option to either use default values or input custom details for email, password, first name, last name, and website.

**Dependency-Check**: It checks for the existence of n98-magerun2.phar and downloads it if not found, ensuring the necessary tool for Magento CLI operations is available, n98 is used for creating customer accounts.

**User Interaction:** The script prompts the user to choose between creating an admin or a customer account and provides prompts for further customization if required.

How added features work: 

[Screencast from 2024-02-25 01-37-40.webm](https://github.com/markshust/docker-magento/assets/43544955/a0234b68-02aa-47d3-bc7f-bf3a1778e4b6)

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
...
+      echo "Email: ${EMAIL}"
+      echo "Firstname: ${FIRSTNAME}"
+      echo "Lastname: ${LASTNAME}"
+      echo "Password: ${PASSWORD}"
+  else
+      echo "Error creating customer account."
+  fi
+}
+
+# Check if n98-magerun2.phar exists, if not download and install
```
markshust, line 51: We don't need lines 51-57 as they are already covered in the `bin/n98-magerun2` script https://github.com/markshust/docker-magento/blob/master/compose/bin/n98-magerun2

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+
+# Function to create admin user
+create_admin_user() {
```
markshust, line 4: It doesn't make sense to me to create a generic, pre-filled admin account here. Let's always ask the user for this data instead.

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
...
+
+  bin/cliq chmod +x n98-magerun2.phar
+  bin/cliq mkdir -p bin
+  bin/cliq mv n98-magerun2.phar bin
+fi
+
+# Prompt user for type of account to create
+read -r -p "Do you want to create an admin (A) or a customer (C) account? [A/C]: " account_type
+
+if [[ "$account_type" == "A" || "$account_type" == "a" ]]; then
```
markshust, line 67: This can be simplified to:
`if [[ "$account_type" == [Aa] ]]; then`

Let's also apply this same logic to the rest of the code in this conditional.

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
...
+    create_admin_user
+elif [[ "$account_type" == "C" || "$account_type" == "c" ]]; then
+    read -r -p "Do you want to use default values for customer account? (Y/N): " use_defaults
```
markshust, line 70: Again, let's just remove this part and always prompt the user to enter this data. This will lead to a more predictable/expected outcome for the user.

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
...
+    if [[ "$use_defaults" == "Y" || "$use_defaults" == "y" ]]; then
+        create_customer_account
+    elif [[ "$use_defaults" == "N" || "$use_defaults" == "n" ]]; then
+        read -r -p "Enter customer email: " EMAIL
+        read -r -p "Enter customer password: " PASSWORD
+        read -r -p "Enter customer firstname: " FIRSTNAME
+        read -r -p "Enter customer lastname: " LASTNAME
+        read -r -p "Enter customer website: " WEBSITE
```
markshust, line 78: It's not apparent that we are asking for a website id or a website code here, so I think we should update this prompt to "Website ID:".

compose/bin/create-user
```diff
@@ -0,0 +1,86 @@
...
+read -r -p "Do you want to create an admin (A) or a customer (C) account? [A/C]: " account_type
+
+if [[ "$account_type" == "A" || "$account_type" == "a" ]]; then
+    create_admin_user
+elif [[ "$account_type" == "C" || "$account_type" == "c" ]]; then
+    read -r -p "Do you want to use default values for customer account? (Y/N): " use_defaults
+    if [[ "$use_defaults" == "Y" || "$use_defaults" == "y" ]]; then
+        create_customer_account
+    elif [[ "$use_defaults" == "N" || "$use_defaults" == "n" ]]; then
+        read -r -p "Enter customer email: " EMAIL
```
markshust, line 74: Let's remove "Enter customer" from all of these prompts to simplify this, and to make it apply to both admins and customers. So this can just say "Email:"

markshust (changes requested):
Thanks for the updates! But I'd like this script simplified, as there is too much going on in the code and it can be made more intuitive with the suggested updates.

YevhenZvieriev:
Hi, @markshust 

I took into account all your amendments and simplified the script.
Now it's four times shorter, but it works as well as before.
Feel free to check it!

markshust (approved):

markshust:
@YevhenZvieriev I feel as though the script was still way too unnecessarily complex.

I think you removed something that shouldn't have been removed (which is the standardization of inputs between the two accounts). It'd be nice for both of these scripts to share the same prompts for email, password, etc. rather than the disconnect between the Magento and magerun scripts.

Also, there were unneeded references to n98-magerun2. Again, all of the functionality to download, install, check n98-magerun2 is in the https://github.com/markshust/docker-magento/blob/master/compose/bin/n98-magerun2 file, so we don't need it anywhere else. There were also same malformed indentations which I removed.

All code in this project strives for elegance + simplicity -- if it doesn't match either of those two aspects, it will always be denied, hung up for a long while, or closed out and not merged.

I updated this PR, and appreciate your initial one! But as a case-study for future PR's, I'd compare:
- The initial PR: https://github.com/markshust/docker-magento/blob/9d1940da4e2e1c6cedf944c8f95c33141d3e724d/compose/bin/create-user
- The final PR: https://github.com/markshust/docker-magento/blob/b6521c303b449bbacbf5e4319af9de8d7220a4cd/compose/bin/create-user

This is 85 lines vs 21 lines. This is less code to maintain, it's easier to read the file and understand what is going on, and it's very elegant -- it doesn't do anything that it doesn't need to do. But, it does just enough to add some extra functionality that goes above and beyond what someone would expect.

markshust:
Every helper script created also needs a corresponding addition to the README, describing the script, and also added to the Makefile! (something that I routinely forget 😅) https://github.com/markshust/docker-magento/pull/1040/commits/8cad6b6aea106d4d750c5ba25ec9dc314d246995

Issue #760, opened 2022-08-31
Cron unexpected fails to start after being enabled


cornernote:
I saw this was asked in #691, however it's not really a permenant start.

I added `restart: always` to the container settings in `docker-compose.yml`, but cron doesn't restart.

You can reproduce this with:

```
$ bin/cron status
cron is running.
$ bin/restart && bin/cron status
cron is not running ... failed!
```

How can I get it to start whenever the container starts?

markshust:
Sorry, I tried everything here but I'm very stuck in figuring this out. Tagging this as help needed.

ssx:
@cornernote are you working on a commerce project?

apresources:
Maybe you could incorporate the idea in this repo https://github.com/markhilton/docker-php-fpm ?

The main php-fpm container calls a script called docker-php-entrypoint, into which is inserted a script which then runs whatever scripts are required. Amongst these is a script that sets up the crontab and starts cron.

markshust:
Thanks for the suggestion. I looked at the PHP 8.1 image and it doesn't look like it has a different `CMD` command other than the default, but perhaps I didn't look into it in enough detail. I'm willing to take a PR for something like this as long as it doesn't overly complicate the bootstrap process of the entry PHP script.

apresources:
> Thanks for the suggestion. I looked at the PHP 8.1 image and it doesn't look like it has a different `CMD` command other than the default, but perhaps I didn't look into it in enough detail. I'm willing to take a PR for something like this as long as it doesn't overly complicate the bootstrap process of the entry PHP script.

Before looking at this, I was working on doing logrotate for our docker-m2 site, which I am running in a separate alpine container with the var/log directory mounted. That container runs the crond as it's CMD. Was wondering if that could be an alternative approach, i.e. run cron on a separate container with the same deps/volumes as php-fpm? Any thoughts?

apresources:
I have got this working but would not suggest making it part of the repo as it gives the app user sudo and messes with the php entrypoint.

I created `images/phpfpm` directory and added a `Dockerfile`

```
FROM markoshust/magento-php:8.1-fpm-1

USER root

RUN apt-get update && apt-get -y install sudo

COPY ./docker-php-entrypoint /usr/local/bin
RUN chmod +x /usr/local/bin/docker-php-entrypoint

RUN usermod -aG sudo app

RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers

USER app
```

[...]

PR #1041, opened 2024-01-11
Implement persistent/reliable cron service using supervisor.

jasonhildebrand:
This PR addresses #760.

There were several non-obvious challenges which I want to document here.
  * if you want to start multiple processes in a single docker container (which IMO is a valid thing to do if they are tightly coupled), then you need a way to start/restart these process.  I'm using supervisor for this purpose.
  * cron needs to run as root, which means that the container entrypoint also needs to run as root (we want to avoid needing sudo to launch cron). the implication of this change is that `docker-compose exec phpfpm` runs as root by default (not app), which sucks.  thankfully, we can update the bin/cli and bin/clinotty wrappers to use the `app` user and maintain existing behaviour (and I have done so).
  * /dev/stdout is inaccessible by an unprivileged user in a container started as root (https://github.com/moby/moby/issues/31243), which makes it difficult to have php-fpm (running as app) to log to the docker console.  The workaround is to set `tty: true` in compose.yaml and add `app` to the `tty` group.

Other notes
  * `bin/cron` has been updated and can still be used to start/stop cron or check status
  * you can set START_CRON=true (or =false) in `env/phpfpm.env` to control whether you want cron to be started when the container is started.

Alternative approaches:
  * My aim was to get cron working in the same container as php-fpm.  An alternative approach would be to run cron in a separate container.  This would require ensuring that the cron container has access to the identical codebase as the php-fpm container (e.g. the same volumes and/or bind mounts), because cron needs to be able to execute php code.  My sense was that keeping this in sync could be tricky, which is one of the reasons I didn't pursue this approach.  However, it would avoid the need for supervisor and would side-step the /dev/stdout permissions issue.
  * Use https://github.com/aptible/supercronic instead of Vixie cron.  supercronic can run as non-root, so that would simplify a number of things. But I have no direct experience with it, so I don't know about compatibility issues.  A quick glance at the issue queue suggests it might not yet be quite mature enough (as of Jan 2024), but looks promising.

At the moment, cron support is only implemented in the PHP 8.2 image.  If this approach is acceptable, I would be happy to extend it to support the PHP 8.1 image as well.


markshust:
Thanks for this PR! Since this is one of the larger updates that came in, I need to test it a little more extensively. I really appreciate the detailed comments, explanations, and reasonings behind this PR.

Since I'd like to get a tagged version of docker-magento out imminently, I'm going to skip over this PR and come back to it after the other PRs are resolved, as this one is already having a conflict or two from the other merged PRs. Once those are all resolved, I'll take a deeper look into this one 👍

markshust:
I decided not to proceed with this update just because it touched too many files, removed some USER dir's, changed TTY settings, etc. — it was too invasive. There are so many places that the TTY settings need to be exactly correct, given automated installs, manual CLI commands, etc. as well as the user settings, I didn't want to make this big of a change to the project. I understand this will not work on prod, but this project is really meant for dev, and I really don't want to support prod at this point in time (it would likely require a far different setup with Swarm/Kubernetes for it to scale properly and be secure & performant). Feel free to fork & pull if you need this though!

PR #1050, opened 2024-01-26
Added instructions for Linux install to README

kmilodenisglez:
added some dependencies that were needed to be able to run magento-docker successfully on a VPS with Debian GNU/Linux 12


kmilodenisglez:
Updated the README.md with some dependencies that needed to be installed on a VPS with Debian GNU/Linux 12 installed.

Dependencies:

* curl 
* libnss3-tools 
* unzip 
* rsync

markshust:
Thanks for the PR. I think it would make sense to move this to the Linux section at https://github.com/markshust/docker-magento#linux instead?

kmilodenisglez:
> Thanks for the PR. I think it would make sense to move this to the Linux section at https://github.com/markshust/docker-magento#linux instead?
> 
> Also, there is only one item here so I don't think there is a need for a 1. bullet point.

yes you're right

kmilodenisglez:
@markshust I just modified the readme.md of this same PR, check it and tell me.

markshust:
@kmilodenisglez for some reason I cannot make changes to this PR myself, so can you make a couple more updates?

1. Move the block that added to just before the `#### The host.docker.internal hostname` line. The lines that followed your previous line block are basic lines about running Docker on Linux, and should come first.
2. Add a description after your heading to remain consistent with the rest of the documentation. It can say something like "docker-magento requires a few system dependencies to be installed on Linux in order to function properly. Please execute the following line from terminal to install these dependencies:"

kmilodenisglez:
> @kmilodenisglez for some reason I cannot make changes to this PR myself, so can you make a couple more updates?
> 
> 1. Move the block that added to just before the `#### The host.docker.internal hostname` line. The lines that followed your previous line block are basic lines about running Docker on Linux, and should come first.
> 2. Add a description after your heading to remain consistent with the rest of the documentation. It can say something like "docker-magento requires a few system dependencies to be installed on Linux in order to function properly. Please execute the following line from terminal to install these dependencies:"

I already updated the readme

PR #1083, opened 2024-03-06
 Switch the script `bin/docker-stats` to the simplified output

### LINKED ISSUE #1075: Error response from daemon:  page not found during executing `bin/docker-stats` script
### Description

I noticed, that after this [commit](https://github.com/markshust/docker-magento/pull/1017/commits/b7418753bebbe7aeaf70b35bffc025480733b435) I got the error  `Error response from daemon:  page not found` during executing `bin/docker-stats`

I got this issue only once, and now, I cannot replicate this, and the script works for me.

Could anyone please confirm the issue? Then I'll fix this.

YevhenZvieriev:
This script provides a simple way to monitor Docker container resource usage using the docker stats command. It continuously displays the CPU and memory usage of active Docker containers in a formatted table and fixes #1075 .

How it works:

Screenshot from 2024-03-06 23-08-56


markshust:
Thanks -- looks good. I just updated the shebang as we use a more portable one within all of the other scripts. Merged in, appreciate it.

PR #1090, opened 2024-03-13
Pin Docker Compose v2.6.6

Stoyvo:
Magento's composer plugin hasn't been updated in over a year, they merged a composer 7 fix but haven't released it... So we'll pin the version for now.

https://github.com/magento/composer-root-update-plugin/issues/40


markshust:
Thank you -- I believe we should be pinning down all versions everywhere, anyways. Merging in 👍

markshust:
FYI, note that any changes made to one PHP version should almost always be made to all of the related docker images. I updated this so it affects all versions.

PR #1093, opened 2024-03-15
Peg to the supported version of Composer

### LINKED ISSUE #1084: Latest Composer is incompatible with latest Magento
**Description**
When setting docker-magento up for the first time, it installs the latest version of Composer from what I see here https://github.com/markshust/docker-magento/blob/69f29dea57f12e8f7363e483b753a299d15cb030/images/php/8.2/Dockerfile#L95. Currently, this installs Composer 2.7.1. After setup, if I run `bin/composer` (without arguments), it outputs an error (below). This seems to be caused by the issue discussed here https://github.com/composer/composer/issues/11843, caused by an incompatibility with Composer 2.7+ and Magento.

**Expected Result**
I think it would be nice if instead of installing the latest version of Composer, it installed the latest version of Composer that's compatible with Magento as outlined here: https://experienceleague.adobe.com/docs/commerce-operations/installation-guide/system-requirements.html. I know that you can technically install different versions of Magento in docker-magento though, so maybe it just defaults to whatever version of Composer is compatible with the default version of Magento that's installed.

**Actual Result**
```
~/magento.test$ bin/composer
Fatal error: Declaration of Magento\ComposerRootUpdatePlugin\Plugin\Commands\OverrideRequireCommand::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output) must be compatible with Composer\Command\RequireCommand::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output): int in /var/www/html/vendor/magento/composer-root-update-plugin/Plugin/Commands/OverrideRequireCommand.php on line 173
```

YevhenZvieriev:
Fix #1084

### DIFF
--- a/images/php/8.2/Dockerfile
+++ b/images/php/8.2/Dockerfile
@@ -84,7 +84,7 @@ RUN version=$(php -r "echo PHP_MAJOR_VERSION.PHP_MINOR_VERSION;") \
     && rm -rf /tmp/blackfire /tmp/blackfire-probe.tar.gz
 
 RUN curl -sS https://getcomposer.org/installer | \
-  php -- --install-dir=/usr/local/bin --filename=composer
+  php -- --install-dir=/usr/local/bin --filename=composer --version=2.2.0
 
 COPY conf/blackfire.ini $PHP_INI_DIR/conf.d/blackfire.ini
 COPY conf/msmtprc /etc/msmtprc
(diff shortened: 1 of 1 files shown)


markshust:
This PR already came in and was merged https://github.com/markshust/docker-magento/pull/1090 - so closing this one out.

PR #1112, opened 2024-04-09
Update supported versions in the `bin/check-dependencies` script

YevhenZvieriev:
Magento 2.4.7 has been released, so I removed the dependencies for such versions as `2.4.7-beta3` and `2.4.7-beta2` from the `bin/check-dependencies` script and added `2.4.7` and `2.4.6-p5` versions.

Screenshot from 2024-04-09 22-08-01


markshust:
Thanks! Note that there are also new versions `2.4.5-p7` and `2.4.4-p8`. Do these versions need related updates?

YevhenZvieriev:
Good point! I'll add the system requirements of the new versions in the next commit.

YevhenZvieriev:
I added system requirements for new versions of Magento 2 so now the script provides system requirements for all supported versions.

Additionally, I added a link to the Magento system requirements documentation from DevDocs if the version entered by the user is not found in the list. This enhancement provides users with easy access to detailed information about the system requirements for running Magento, facilitating smoother installation and setup processes.

Screenshot from 2024-04-13 13-30-43

PR #1125, opened 2024-04-17
Removes obsolete `version` directive from docker compose files

michaellehmkuhl:
Latest version of docker compose started throwing a **WARN[0000]** notification about `version` being obsolete.

https://github.com/docker/compose/issues/11628

### DIFF
--- a/compose/compose.dev-linux.yaml
+++ b/compose/compose.dev-linux.yaml
@@ -1,5 +1,3 @@
-version: "3"
-
 services:
   app:
     volumes: &appvolumes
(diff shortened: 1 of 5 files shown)


markshust:
Ah, I didn't know this! Thanks for this @michaellehmkuhl.

I am updating how PR's are merged. Rather than merging this directly to `master` (which would affect all live installs immediately), I'd like to instead merge into a new branch named `release/next`. This way, we can stage updates, and have predictable release versions.

@michaellehmkuhl can you update this PR so that the target branch is `release/next` instead of `master`?

markshust:
Never mind, figured out how to do this for existing PR's!

PR #1131, opened 2024-04-18
Update Integration testing configuration to use OpenSearch

TuVanDev:
This pull request updates the Integration testing configuration to use OpenSearch, as we have moved from Elasticsearch to OpenSearch.


TuVanDev:
I also added the `consumers-wait-for-messages` config to the Integration testing configuration as the change was 
introduced in Magento 2.4.5, which mentioned fixing the issue:

> ACP2E-257: error: : Future Update already exists in this time range. Set a different range and try again when updating end date

Related commit in the Magento codebase: https://github.com/magento/magento2/commit/eb756fc9e6ca207fd3c48343714c0e015016ff97

PR #1136, opened 2024-04-24
Fix SSL cert generation when domain has a port included

supercid:
This fixes the case where someone tries to run this setup with a domain name that includes a port number

For example, running `bin/setup magento.test:8443`will cause the certificate to fail with an invalid domain name error, since the port should no be included. However, the port should be used in other parts of `bin/setup domain`


markshust:
Very cool, thanks for this!

PR #1143, opened 2024-04-26
Moved check for empty directory to top of scripts

YevhenZvieriev:
In this pull request, a check has been moved to the top of the scripts to verify the existence of the `bin` and `src` directories BEFORE taking any actions. This was done to ensure that the directory is empty BEFORE beginning the process, and prevents unintentional overwriting of content, as well as any errors that might occur during execution.


markshust:
Thanks, this definitely makes sense 👍

PR #1163, opened 2024-05-27
Add rsync to PHP images for Deployer support

gbobts:
Hello,
Would you consider adding rsync ?

### DIFF
--- a/images/php/8.1/Dockerfile
+++ b/images/php/8.1/Dockerfile
@@ -34,6 +34,7 @@ RUN apt-get update && apt-get install -y \
     msmtp \
     nodejs \
     procps \
+    rsync \
     strace \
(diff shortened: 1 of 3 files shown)


akosglue:
"potentially benefiting the workflows that use this image" -> an example is deployer's upload: https://github.com/deployphp/deployer/blob/682ff2171d115bf699839852079832ad26534b8e/src/functions.php#L582

(when building magento with this image and using artifact deploy)

markshust:
@gbobts thanks for the link and mention of deployer, that sold me. I also made the update to the PHP 8.1 and 8.2 images. I'll be rebuilding all of these images very shortly in a new build, as it's been a while since they had an update. Appreciate it!

PR #1177, opened 2024-06-23
phpMyAdmin for linux

YevhenZvieriev:
Accordingly, to [install instructions for Linux](https://github.com/markshust/docker-magento?tab=readme-ov-file#linux) we should copy `compose.dev-linux.yaml` to `compose.dev.yaml` before installing Magento but the `compose.dev-linux.yaml` doesn`t contain phpMyAdmin, this PR adds it.


markshust (approved):

markshust:
Thank you, pulling this one in!

PR #1171, opened 2024-06-07
Add `bin/ece-patches` command

DmitryFurs:
## Summary

This PR adds a new command `bin/ece-patches` to enable running Adobe Commerce Cloud patches tool commands from the host inside the container. This addition aims to simplify the process of applying patches by providing a straightforward command.

## Usage

To use the new alias, run the following command from the host:
```shell
bin/ece-patches apply
```
image

This command will execute the `./vendor/bin/ece-patches` inside the container, allowing you to apply patches as documented in the [Adobe Commerce Cloud Service documentation](https://experienceleague.adobe.com/en/docs/commerce-cloud-service/user-guide/develop/upgrade/apply-patches#apply-a-patch-in-a-local-environment).

PR #1185, opened 2024-07-05
Create ece-patches bin command to apply patches

ejaz-alam:


compose/bin/ece-patches
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if ! bin/cliq ls vendor/bin/ece-patches; then
+    echo "ece-patches tool is not installed, please reinstall."
+fi
+
+bin/cli chmod +x vendor/bin/ece-patches
```
piotrkwiecinski, line 6: We shouldn't have to run chmod each time.
If user has invalid permissions they should run this command as a one off.

compose/bin/ece-patches
```diff
@@ -0,0 +1,7 @@
...
+bin/cli /app/vendor/bin/ece-patches "$@"
```
piotrkwiecinski, line 7: This path should be `vendor/bin/ece-patches` right? The same as above.

DmitryFurs:
Hi @ejaz-alam I have already created such a pr https://github.com/markshust/docker-magento/pull/1171

PR #1195, opened 2024-07-19
Update n98-magerun2 CLI to use a single insert with column names per row

TuVanDev:
Description: Update `n98-magerun2` CLI to use a single insert with column names per row. This change will facilitate easier data searching through the dumped database file.

Details: By default, the `mysqldump` command generates `INSERT` query values on new lines, which complicates data searching through the dumped database file.
For example:
```
zgrep -P "VALUES\s*\(.*\d:\d:...*'" database.sql | awk '{print $3}' | uniq
```
Adding the `--human-readable` option to magerun will enforce a single insert with column names per row (equivalent to `--extended-insert=FALSE` in the `mysqldump` command).

Current:
`bin/cli bin/n98-magerun2.phar "$@"`

Proposed change:
`bin/cli bin/n98-magerun2.phar --human-readable "$@"`

### DIFF
--- a/compose/bin/n98-magerun2
+++ b/compose/bin/n98-magerun2
@@ -11,4 +11,4 @@ if ! bin/cliq ls bin/n98-magerun2.phar; then
   bin/cliq mv n98-magerun2.phar bin
 fi
 
-bin/cli bin/n98-magerun2.phar "$@"
+bin/cli bin/n98-magerun2.phar --human-readable "$@"


markshust (approved):

piotrkwiecinski:
@markshust @TuVanDev if it affects db dump shouldn't it in https://github.com/markshust/docker-magento/blob/release/next/compose/bin/mysqldump#L2 instead of all magerun commands?

n98-magerun2 should be just a bypass to a tool itself

TuVanDev:
Great catch! Thanks @piotrkwiecinski.  I appreciate it.

I added it to `bin/mysqldump` while resolving the [related issue](https://github.com/genecommerce/module-encryption-key-manager/?tab=readme-ov-file#fully-rotate-your-old-keys) in my project. However, when I went to open a new pull request to the repository, I accidentally made the change to `bin/n98-magenrun2` 😅 .

PR #1200, opened 2024-07-29
fix: Updated bin/setup-ssl to support multi site ssl generation

nathanchick:
This PR re-adds support for multiple domain SSL generation while maintaining stripping ports from each domain.


ssx (approved):

ssx:
@markshust can we get some eyes on this? currently breaking out workflows internally.

markshust (approved):

PR #1210, opened 2024-08-21
Add PHP FTP ext on Docker setup

rafaelstz:
PHP >= 8.2.15 removed the FTP extension, but Magento still requests its use for FTP configurations. So, I'm adding it to the Docker PHP 8.2 and 8.3 for new builds and upgrades.

Reference: https://github.com/magento/magento2/issues/39083


markshust:
Appreciate it! It's been a while since I've tagged a new release, so I'll rebuild the images and do this soon.

PR #1242, opened 2024-11-09
Update system requirements helper script overhauled to make it easier to use, read and update. Updated requirements for Magento 2.4.

YevhenZvieriev:
This PR adds the system requirements of Magento 2.4 (2.4.4-p10, 2.4.4-p11, 2.4.5-p10, 2.4.5-p9, 2.4.6-p8, 2.4.6-p7, 2.4.7-p3, 2.4.7-p2). 

It also adds an important MySQL compatibility notice: A note has been added to indicate that Magento versions 2.4.4 and 2.4.5 support MySQL versions 8.0.0 to 8.0.28 only.


markshust:
This changeset is so hard to read. We need to change the format of it, so we know what version goes with what set of system requirements.

markshust (approved):

PR #1352, opened 2025-06-19
Fix OpenSearch failing to start on Linux

### LINKED ISSUE #1346: dependency failed to start: container 248-opensearch-1 is unhealthy
Image

Operating System: Ubuntu 24
Magento: 2.4.8
Installation Execution: Manual Magento installation Using Docker

How to resolve this error can anyone guide me?

mrifqyabdallah:
### 1. Increase the retries number on the `compose.healthcheck.yaml` file, from 3 to 30

Fixes #1346. See [my comment](https://github.com/markshust/docker-magento/issues/1346#issuecomment-2986531014) inside.

### 2. Fixed typo and syntax error on the `Makefile`

After cloning the repo and running `make` on my CLI, I get the following error:

Screenshot 2025-06-19 135157

At line 42 of the Makefile file, turns out there's a missing parentheses. 

I also escape all the backticks and change all comma signs with the `$(comma)` variable in order to remove the `unexpected EOF` warning.


markshust:
Fixes https://github.com/markshust/docker-magento/issues/1346

markshust:
Thanks so much for the PR and extra fixes.

I did decrease the retry count from 30 though to 6. The default retry interval was `5s`, which means that when it was set to `3`, if a service did not start within 15 seconds it would fail. Setting this to `30` would mean that we would need to wait... 150 seconds! Which is nearly 3 minutes. So instead, I changed this to `6`, which would increase the timeout from 15 seconds to 30 seconds.

Note that this was a global change made to all services, so we need to be mindful of the updates here. We actually do want the Docker startup to fail if things take too long to get up & running, so users aren't twiddling their thumbs for a few minutes when an issue is preventing the start. I'm guessing that `6` will work just fine, but if it doesn't, we will probably need another fix, or a setting that targets the `opensearch` service specifically.

I'm merging this in as a bugfix release in tag `52.0.2` and it'll be out momentarily.

PR #1355, opened 2025-07-06
feat: add separate phpfpm container dedicated to xdebug

### LINKED ISSUE #1230: Change xdebug
**Description**
Today, you have to manually enable/disable xdebug, which is taking time, and sometime the user can forget to enable it and wonders why the breakpoint didnt trigger, or, you have the option to always have it enabled, which is negatively affecting performance in local. 

It would be great if we could have the same logic in place as what today adobe commerce cloud has, where, even in production i can trigger xdebug, without it affecting regular users
https://experienceleague.adobe.com/en/docs/commerce-cloud-service/user-guide/develop/test/debug#debug-for-pro-staging-and-production
image
What we could try to do, implement something similar to comment #2 from 
https://github.com/magento/magento-cloud-docker/issues/112#issuecomment-565345812

**Steps To Reproduce**
1. N/A

**Expected Result**
* xdebug triggers only when XDEBUG_SESSION cookie present in request. This request is being sent to a separate process with loaded and configured xdebug extension
* if XDEBUG_SESSION cookie not present, request is sent to a process which does not have xdebug loaded at all

**Actual Result**
xdebug is either enabled or disabled (but, even if enabled or disabled, the extension is still loaded)

jarzebowsky:
⚠️ The PR is not finished as it needs some answers regarding the process of publishing the images within the changes.
It should probably also include the changes to the README (didn't do anything as the initial approve should be happening before that happens).

⚠️ The changes are applied only to the default images and should be expanded to the rest of them after initial approval of the solution.

There is also one thing that I don't have idea how to solve as it requires manual change after setting up the project - perhaps some core contributors have better understanding of the repository to drop the lead or give concrete approach for that.

## Context

This resolves #1230 which drops manual interaction with turning on/off xdebug which is problematic to remember for a user but also even if it's turned off the xdebug lib is all the time installed.

## Description

We need to add mapping to nginx configuration so it will pass the request to concrete phpfpm container based on the cookie attached to request.

## Changes in the codebase

The changes includes separation of `phpfpm` container to two separate container (`phpfpm`, `phpfpm-xdebug`). This allows us to drop xdebug for `phpfpm` container.
We also have changed the socket names for each of the containers which are named accordingly `/sock/phpfpm.sock` and `/sock/phpfpm-xdebug.sock`. Since we are running both in the same time there were conflicts regarding names so that what was needed to be done.

There are also changes regarding nginx configuration regarding `app` container which uses this image. First - we adjusted upstream definition after changes of name for currently existing phpfpm one and added second one which covers new one regarding `xdebug`.

The magic and thing that does the job is the next one: **map**. Based on the cookie being sent within the request `XDEBUG_SESSION` we define to which phpfpm container the request should be passed.

## Additional information

The changes were tested within VSCode and PHPStorm configuration and we would need to adjust the README regarding the usage of the xdebug thing. For PHPStorm we need to declare the concrete name for record in `PHP | Servers` so it will map correctly to the files tree.

This is a part of the changes that should be applied because the same thing should happen to CLI usage so instead of using whole FPM to run commands we could use smaller php CLI and php CLI with xdebug. The final outcome should give us:
* phpfpm
* phpfpm-xdebug
* php
* php-xdebugThe changes contains:
- drop xdebug from current php-fpm container
- add new php-fpm container with xdebug configuration
- changes to nginx to name sockets accordingly and apply mapping to do it based on cookie in request


compose/compose.yaml
```diff
@@ -12,7 +12,10 @@
 
 services:
   app:
-    image: markoshust/magento-nginx:1.24-0
+    # image: markoshust/magento-nginx:1.24-0
+    build:
+      context: ./images/nginx/1.24
+      dockerfile: Dockerfile
```
jarzebowsky: @markshust This should be replaced with dedicated image build

compose/compose.yaml
```diff
@@ -30,11 +33,25 @@ services:
       #- "host.docker.internal:host-gateway"
 
   phpfpm:
-    image: markoshust/magento-php:8.3-fpm-4
+    # image: markoshust/magento-php:8.3-fpm-4
+    build:
+      context: ./images/php/8.3
+      dockerfile: Dockerfile
```
jarzebowsky: @markshust This should be replaced with dedicated image build

compose/compose.yaml
```diff
@@ -30,11 +33,25 @@ services:
...
     volumes: *appvolumes
     env_file: env/phpfpm.env
     #extra_hosts: *appextrahosts
 
+  phpfpm-xdebug:
+    # image: markoshust/magento-php:8.3-fpm-4
+    build:
+      context: ./images/php/8.3-xdebug
+      dockerfile: Dockerfile
```
jarzebowsky: @markshust This should be replaced with dedicated image build

images/nginx/1.24/conf/default.conf
```diff
@@ -1,5 +1,14 @@
...
-  server unix:/sock/docker.sock;
+upstream fastcgi_phpfpm {
+  server unix:/sock/phpfpm.sock;
+}
+
+upstream fastcgi_phpfpm_xdebug {
+  server unix:/sock/phpfpm-xdebug.sock;
+}
+
+map $cookie_XDEBUG_SESSION $fastcgi_backend {
```
jarzebowsky, line 9: The problem I mentioned in the description is that mapping requires to put `$` sign when assigned to variable. There is a logic which copies `nginx.conf.sample` to `nginx.conf` but this sample file has 
```
fastcgi_pass   fastcgi_backend;
```

but it should be 
```
fastcgi_pass   $fastcgi_backend;
```

The occurence is repeated 3 times - perhaps some post-script with `sed` that would replace it?

lib/onelinesetup
```diff
@@ -5,7 +5,8 @@ DOMAIN=${1:-magento.test}
 EDITION=${2}
 VERSION=${3}
 
-curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/template | bash
+# curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/template | bash
+curl -s https://raw.githubusercontent.com/Jarzebowsky/docker-magento/xdebug-change-trigger-method/lib/template | bash
```
jarzebowsky: This change was introduced to fully test the approach so it should be reverted. 
It's left for the sake of testing it locally.

lib/template
```diff
@@ -6,13 +6,16 @@ if [ -d "./bin" ]; then
...
+git remote add origin https://github.com/Jarzebowsky/docker-magento
 git fetch origin -qqq
-git checkout origin/master -- compose
+# git checkout origin/master -- compose
+git checkout origin/xdebug-change-trigger-method -- compose images
 
 mv compose/* ./
 mv compose/.gitignore ./
 mv compose/.vscode ./
+mv images/* ./images
```
jarzebowsky: This change was introduced to fully test the approach so it should be reverted. 
It's left for the sake of testing it locally.

markshust:
Heads up — I pushed a bunch of significant changes to this branch:

- Merged `release/next` back in (resolved one debug-cli conflict)
- Refactored all PHP images (8.1, 8.2, 8.3, 8.4) into multi-stage Dockerfiles (`base` + `xdebug` targets) sharing a single set of config files; deleted the standalone `images/php/8.3-xdebug/` directory
- Updated CI workflows to build and publish both targets, with paired revision numbers (e.g. `8.3-fpm-6` and `8.3-fpm-xdebug-6`)
- Bumped base image revisions across the board since the new base no longer includes Xdebug (behavior change)
- Bumped nginx image to `1.24-1` with the updated `default.conf` (two upstreams + cookie map)
- Added `bin/setup-nginx` that rewrites Magento's `nginx.conf` to use `$fastcgi_backend` so cookie routing works out of the box; auto-invoked by `bin/setup`
- `bin/xdebug` replaced with a deprecation stub
- `bin/debug-cli` simplified to just exec into `phpfpm-xdebug`
- `bin/test/unit-xdebug` and `bin/test/unit-coverage` now route through `bin/debug-cli`
- Reverted the `Jarzebowsky/` fork references in `lib/onelinesetup` and `lib/template`
- Reverted compose.yaml back to published `image:` references (no `build:` blocks)
- Updated README, CHANGELOG, and Makefile help text
- Retargeted the PR base from `master` to `release/next`

This branch still needs my manual review and end-to-end local verification before merging. Don't merge yet.

PR #1364, opened 2025-08-12
Update compose.yaml

Niyasp:
getting port conflict on my system


markshust:
This PR is not safe to merge. The default ports are what is required for this project. We may update this in the future to support multiple sites running at once, but that will require a much more involved PR.

PR #1374, opened 2025-09-05
Fix: Prevent CDPATH from corrupting directory paths in copy scripts

CasEbb:
This PR suppresses output from the `cd` command when setting REAL_SRC in `copyfromcontainer` and `copytocontainer`. This prevents a bug where REAL_SRC is filled with more than one path when the user makes use of a $CDPATH environment variable. [Per `cd`'s man](https://man7.org/linux/man-pages/man1/cd.1p.html#STDOUT):

> If a non-empty directory name from CDPATH is used [...] an absolute pathname of the new working directory shall be writen to the standard output as follows:
> 
>     "%s\n", <new directory>

Redirecting `cd`'s output to `/dev/null` makes sure the variable is only filled with the output from `pwd`.

### DIFF
--- a/compose/bin/copyfromcontainer
+++ b/compose/bin/copyfromcontainer
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 [ -z "$1" ] && echo "Please specify a directory or file to copy from container (ex. vendor, --all)" && exit
 
-REAL_SRC=$(cd -P "src" && pwd)
+REAL_SRC=$(cd -P "src" >/dev/null && pwd)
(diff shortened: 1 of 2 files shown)


CasEbb:
Fun fact: I'm using a CDPATH and the current version copies files to a dir named "src\n".

```sh
$ ls -lb | grep src
drwxr-xr-x  34 cas  staff  1088  5 sep. 10:39 src
drwxr-xr-x   3 cas  staff    96  5 sep. 10:40 src\n
```

markshust:
Thanks, merged! It'll be included in the next release.

PR #1385, opened 2025-10-12
Fix: avoid 'Warning: magento.env file not found' when running bin/setup

angelvilaplana:
### Summary

Previously, the setup script only checked for `../env/magento.env`.  
As a result, when the `magento.env` file was located in `./env`, the script showed the warning:

image

Also, when the env file was not found, the admin credentials did not appear in the output:

image

### Changes

- Added a check for `./env/magento.env` before checking `../env/magento.env`.
- This ensures the environment file is sourced correctly.

### Result
- The warning no longer appears when the file exists in the current directory.
- The admin credentials are now displayed correctly after setup:

image

### DIFF
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -o errexit
-if [ -f "../env/magento.env" ]; then
+if [ -f "./env/magento.env" ]; then
+    source "./env/magento.env"
+elif [ -f "../env/magento.env" ]; then
     source "../env/magento.env"
 else
     echo "Warning: magento.env file not found."


markshust:
Looks like shellcheck is failing. Besides that, why would the `env` directory be located at `src/env`? This is a non-standard location.

angelvilaplana:
Hi @markshust, 

Yes, you're right. The `env` folder is in the root directory, and the `src` folder is also in the root but in a different subfolder. However, even in this case, I’m still getting the same error. When I run `bin/setup`, it tries to execute outside the project folder, so the `env` file can’t be found.

Hmm, I see that the test is running `./compose/bin/setup`. In my case, I run `bin/setup` to install the environment. It should still work anyway, since I added a condition to make sure the file `./env/magento.env` exists before running the `source` command.

markshust:
Closing in favor of https://github.com/markshust/docker-magento/pull/1403 that was just merged in

PR #1393, opened 2025-12-12
Disable Admin Password Expiration And Fix Sample Data Upgrade Order

aimanecouissi:
This change disables admin password expiration in the development environment by setting the password lifetime to 0, so admin users are not forced to regularly change their passwords while developing.

It also moves bin/magento setup:upgrade to run immediately after sampledata:deploy. Running the upgrade only at the end cause sample data to not be properly applied and its modules appear as disabled. Running it right after deploying sample data ensures all sample data modules are correctly registered before continuing with the rest of the setup.


markshust:
I didn't see this PR so already fixed the setup:upgrade issue, however I'll pull in your other change. Thanks for this!

PR #1395, opened 2025-12-19
Set executable flag on `bin/docker-start` and `bin/init`

CasEbb:
`bin/docker-start` and `bin/init` are missing `+x` on their file modes so they aren't executable from the command line.

```zsh
$ bin/init
zsh: permission denied: bin/init
```


markshust:
Thanks for this PR -- completely missed this on the previous PR pull!

PR #1397, opened 2026-01-15
feat: support compose.override.yaml files in docker-compose script

BR-RedEnzian:
### why this feature is useful

if you need to change e.g. ports of one or more services, using a compose.override.yaml file is update save.
in my case, i run my own mariadb server on my machine, so port 3306 is assigned to this server and i cant access the containered version of mariadb with DataGrip.

### current behaviour

docker-compose does currently ignore compose.override.yaml files

### after the changes

this change will add an override file (if it exists) to the docker compose command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,22 @@ We recommend keeping your docker config files in version control, so you can mon
 
 It is recommended to keep your root docker config files in one repository, and your Magento code setup in another. This ensures the Magento base path lives at the top of one specific repository, which makes automated build pipelines and deployments easy to manage, and maintains compatibility with projects such as Magento Cloud.
 
+### Persisting local compose changes across updates
+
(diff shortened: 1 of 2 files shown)


markshust:
Hi @BR-RedEnzian,

Looks like a nice update, however that is kinda what the `compose.dev.yaml` file is for. That file overrides the base `compose.yaml` file, so you should be able to put anything in that file meant to override the base. Did you try that out?

Cheers,
Mark

BR-RedEnzian:
> Hi @BR-RedEnzian,
> 
> Looks like a nice update, however that is kinda what the `compose.dev.yaml` file is for. That file overrides the base `compose.yaml` file, so you should be able to put anything in that file meant to override the base. Did you try that out?
> 
> Cheers, Mark

Hi,

yes i know the .dev file but my problem is/was, that this file is overwritten whenever used the bin/update script.

with the override file, developers have a way to keep their changes for docker compose files even when updating.

markshust:
Thanks for the contribution, @BR-RedEnzian! Your reasoning about `bin/update` overwriting `compose.dev.yaml` is spot on — `compose.override.yaml` is exactly the right way to handle update-safe local customizations, and it also restores Docker Compose's standard convention that the explicit `-f` flags had inadvertently disabled.

I pushed a follow-up commit to your branch with README updates that document the new behavior, including a worked example. Merging now.

PR #1403, opened 2026-01-28
Source env files relative to script path in bin scripts

CasEbb:
When following the [installation instructions](https://github.com/markshust/docker-magento?tab=readme-ov-file#new-projects) from the readme, running `bin/setup` yields a warning `magento.env file not found`. This is because the environment is sourced from `../env/magento.env`, and assuming we're running the script in `$PWD`, it is trying to source the env from `$PWD/../env/magento.env`, while the file is in `$PWD/env/magento.env` (assuming we didn't `cd` into `bin/`).

The original fix changed `bin/setup` to prepend the path with the script dir so it always finds `magento.env` regardless of the working directory. This is also probably a more durable solution than the one proposed in #1385.

## Update — extended for consistency

The same cwd-dependent sourcing pattern existed in `bin/setup-install`, `bin/setup-integration-tests`, and `bin/mysql` (all using `source env/<file>.env`). To keep the bin scripts consistent and avoid the same class of bug surfacing elsewhere, this PR now applies a script-relative `ENV_DIR` to all four scripts:

```bash
ENV_DIR="$(dirname "$0")/../env"
# shellcheck source=../env/db.env
source "$ENV_DIR/db.env"
```

This pattern:
- Works regardless of the user's `cwd` when invoking the script
- Preserves the existing `# shellcheck source=...` directives so ShellCheck CI still follows the env files for full variable lint coverage
- Keeps the same overall structure as the original code — only the path argument changes

Files updated: `compose/bin/setup`, `compose/bin/setup-install`, `compose/bin/setup-integration-tests`, `compose/bin/mysql`.

### DIFF
--- a/compose/bin/mysql
+++ b/compose/bin/mysql
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
+ENV_DIR="$(dirname "$0")/../env"
 # shellcheck source=../env/db.env
-source env/db.env
+source "$ENV_DIR/db.env"
 if [ -t 0 ]; then
   # Need tty to run mysql shell
   bin/cli mysql -h"${MYSQL_HOST}" -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" "${MYSQL_DATABASE}" "$@"
--- a/compose/bin/setup
+++ b/compose/bin/setup
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -o errexit
-if [ -f "../env/magento.env" ]; then
-    source "../env/magento.env"
+if [ -f "$(dirname "$0")/../env/magento.env" ]; then
+    source "$(dirname "$0")/../env/magento.env"
 else
     echo "Warning: magento.env file not found."
 fi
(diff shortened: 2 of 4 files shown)


markshust:
Thanks for the fix @CasEbb — I pushed a follow-up commit that extends the same script-relative env-sourcing pattern to `bin/setup-install`, `bin/setup-integration-tests`, and `bin/mysql`, since all three had the same cwd-dependent `source env/<file>.env` style and would have hit the same class of issue if invoked from outside the project root.

The pattern landed on:

```bash
ENV_DIR="$(dirname "$0")/../env"
# shellcheck source=../env/db.env
source "$ENV_DIR/db.env"
```

I considered a shared `_envsource` helper to be more DRY, but it would have broken the `# shellcheck source=...` directives (ShellCheck doesn't follow indirection through function calls, so SC2154 would have lit up across `setup-install`). The `ENV_DIR` variable keeps things cwd-independent while preserving full ShellCheck coverage in CI.

I also updated the PR title and description to reflect the expanded scope. LMK if anything looks off.

PR #1410, opened 2026-03-02
Refactor copy scripts to reuse resolved container ID

DmitryFurs:
This PR refactors `bin/copyfromcontainer` and `bin/copytocontainer` to resolve the `phpfpm` container ID once and reuse it instead of executing:
```
bin/docker-compose ps -q phpfpm | awk '{print $1}'
```
multiple times.


markshust:
Looks great, thanks for the refactor. Definitely makes things a bit simpler 👍

PR #1412, opened 2026-03-05
Update README.md Composer Auth DevDoc Link

pivitparkour941:
Previous link no longer works. New link directs users to the Commerce Marketplace and should be straight forward from there.


markshust:
Thank you, appreciate it!

PR #1449, opened 2026-08-21
feat: add Varnish support

angelvilaplana:
### Summary

This PR adds Varnish support.

It follows the discussion https://github.com/markshust/docker-magento/pull/441#issuecomment-4500668199.

### Changes

- Added the Varnish 7.7 Docker image and GitHub Actions build.
- Added Varnish and Nginx configuration files.
- Added the Varnish service to Docker Compose.
- Added documentation for enabling and configuring Varnish.

### Result

Varnish can now be easily enabled, making it possible to test Magento with full-page caching closer to a production setup.


markshust:
Thanks for picking this back up, @angelvilaplana — this is really close to exactly the shape I laid out in #441! 🙏

I pushed a small follow-up commit (22740a7) to your branch with a few tweaks rather than bouncing it back:

- The nginx conf is now enabled through a compose `configs:` entry on `app`, since mounting it inside the `&appvolumes` anchor also leaked it into `phpfpm`/`phpfpm-xdebug`
- There's a single VCL now (`template/varnish/default.vcl`), mounted from `compose.yaml`. The duplicate baked into the image and the `compose.dev.yaml` block are gone, so enabling Varnish only touches `compose.yaml` + `compose.healthcheck.yaml`
- Fixed the VCL header comments (filename, and a regenerate command that actually lands the file on the host)
- README now covers removing `http_cache_hosts` from `env.php` when disabling, so Magento stops trying to purge a missing host

With Varnish left commented out, the resolved compose config is identical to `release/next`, so nothing changes for anyone who doesn't opt in.

Thanks again for the contribution!

markshust:
Merged! 🎉 The image is now live on Docker Hub as [`markoshust/magento-varnish`](https://hub.docker.com/r/markoshust/magento-varnish/tags) (`7.7` / `7.7-0`, amd64 + arm64), so enabling Varnish works out of the box. This will go out with the next release.

Thanks again, @angelvilaplana!

PR #1450, opened 2026-08-28
Add Hyvä/Tailwind build commands (hyva-build, hyva-watch, setup-hyva)

hans2103:
Hyvä themes build CSS with Tailwind rather than Grunt/LESS. Their `web/tailwind/hyva.config.json` typically scans `vendor/hyva-themes/...` to generate classes, and since `vendor/` only exists inside the `phpfpm` container (not bind-mounted to the host), the build has to run there.

Adds three commands mirroring the existing Grunt workflow:
- `bin/setup-hyva` — `npm ci` for the theme's Tailwind project
- `bin/hyva-build` — one-off minified build
- `bin/hyva-watch` — rebuild on change

All three auto-detect the theme directory via a new `bin/get-theme` helper, which finds the single `web/tailwind` folder under `src/app/design/frontend` (an explicit theme directory can still be passed as the first argument, e.g. for projects with multiple Hyvä themes).

README updated with reference entries and a new "Hyvä + Tailwind for Frontend Development" section alongside the existing Grunt one.

## Test plan
- [x] `shellcheck -x -e SC2181 -P compose/bin -P compose/env` passes clean on the four new scripts
- [x] Verified end-to-end against a real Hyvä theme (auto-detect, explicit theme-dir override, and the zero/multiple-theme error paths)


markshust:
Thanks for this, @hans2103 — nice, focused addition, and running the build inside `phpfpm` is exactly right given `vendor/` lives there. 🙏

I pushed a small follow-up commit (6f1b92b) and retargeted this to `release/next`:

- `bin/get-theme` now matches `src/app/design/frontend/*/*/web/tailwind` directly instead of an unbounded `find`, which walked the theme's bind-mounted `node_modules` (slow on macOS) and could pick up false matches
- The three commands now use `npm --prefix "<theme>/web/tailwind"` rather than interpolating the path into a nested `bash -c`, and `hyva-watch` goes through `bin/npm` like `bin/grunt` does

Thanks again for the contribution!

